PR #12, opened 2016-05-27
finnish translation updated. added otherRequiredError text to localiz…

topiporkka:
finnish translation updated. added otherRequiredError text to localization


andrewtelnov:
I have missed the text in "question_baseselect.ts" indeed!
Thank you!

PR #29, opened 2016-06-28
Add per question root css attribute

mirmdasif:
Apart from ability of changing the root css in question level possible usage can be indentation of  of question.

Example 
  Q1
  ------Q2
  --------------Q3

### DIFF
--- a/src/knockout/templates/question.html
+++ b/src/knockout/templates/question.html
@@ -1,5 +1,5 @@
 ﻿<script type="text/html" id="survey-question">
-    <div data-bind="visible: question.koVisible(), css: $root.css.question.root, attr: {id: id}">
+    <div data-bind="visible: question.koVisible(), css: $root.css.question.root + ' ' + question.rootCss, attr: {id: id}">
         <!-- ko if: question.hasTitle -->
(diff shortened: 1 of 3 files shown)


andrewtelnov:
Hi,
Aha, I have reviewed the code.
I understand what you are trying to do now. However, I do not like that the model and presentation is mixed now.css - it is a part of presentation and it becomes a part of a model.
I am thinking about implementing the react native in future, and there is no css here. The main idea, that you do not need to change the survey json model, the engine (player) will be able to run it.

andrewtelnov:
I think a better way to do it: to add level or indent property into the question, the default value is 0, possible values are: 0, 1, 2, 3.
And then have indent1, indent2 and indent3 css. If the indent is different from 0, the corresponded css will be added into the question class.
What do you think?

mirmdasif:
I was also thinking about that. Adding a css class to model looks bad and kinda broke MVC philosophy. Definitely adding indent property is much better solution.

mirmdasif:
@andrewtelnov will you add this functionality or should I update my pull request?

andrewtelnov:
It is up to you. I may add it this week or early next week.

andrewtelnov:
https://github.com/andrewtelnov/surveyjs/issues/30

Issue #131, opened 2016-11-25
Using a checkbox with react throws a warning about an uncontrolled input becoming controlled


jdalegonzalez:
This issue is in `SurveyQuestionCheckboxItem::render`.  The code:
```javascript
var isChecked = this.question.value && this.question.value.indexOf(this.item.value) > -1;
```
sets `isChecked` to `undefined`, which means that the input created by `SuveryQuestionCheckboxItem::renderCheckbox` sets the `checked` attribute of the input item it creates to `undefined`.  Having an undefined value for a checked attribute makes react decide that the item is uncontrolled.  A later call to `setState` in `SuveryQuestionCheckboxItem::handleOnChange` makes react believe the input should be controlled and causes the warning to get thrown.

```javascript
var isChecked = t(his.question.value && this.question.value.indexOf(this.item.value) > -1)||false;
```

prevents the warning - I'll issue a pull request for this change.

HOWEVER: I'm not enough of a react guru to know if the value for the checkbox is being handled correctly from a react perspective or not.

PR #132, opened 2016-11-25
Don't set isChecked to undefined if the value element of the question property is undefined.

jdalegonzalez:
This is a proposed fix for issue #131


andrewtelnov:
Thank you!

jdalegonzalez:
Happy to do it.  Thanks for creating and maintaining this library.

PR #134, opened 2016-11-26
Scroll to the top of the page and set focus to the first input in react

### LINKED ISSUE #133: With react, switching pages doesn't scroll the top question into view
This behavior seems to be mostly correct in the test function for the survey designer but in the react version, switching pages doesn't reset the scroll position to reveal the top item.

jdalegonzalez:
Intended to fix #133.  This patch to reactpage.tsx sets focus to the first input and scrolls to the top of the page on page transition.  It's important to note that document elements and the survey title won't be scrolled into view and that behavior might be more desirable.  I chose to scroll the element containing the page to the top rather than the page itself so that the top padding/margin is preserved.

### DIFF
--- a/src/react/reactpage.tsx
+++ b/src/react/reactpage.tsx
@@ -11,6 +11,8 @@ export class SurveyPage extends React.Component<any, any> {
     private survey: SurveyModel;
     private creator: ISurveyCreator;
     protected css: any;
+    protected pageNode: any;
+
     constructor(props: any) {
         super(props);
         this.page = props.page;
@@ -24,6 +26,25 @@ export class SurveyPage extends React.Component<any, any> {
         this.creator = nextProps.creator;
         this.css = nextProps.css;
     }
+    componentDidMount() {
+        if (this.pageNode) {
+            var firstField = this.pageNode.querySelector(
+                "input:not(:disabled):not([readonly]):not([type=hidden])" +  
+                ",select:not(:disabled):not([readonly])"+
+                ",textarea:not(:disabled):not([readonly])");
+            firstField && firstField.focus();
+        }
+    }
+    componentDidUpdate() {
+        if (this.page.isVisible && this.page.isActive && this.pageNode) {
+            var firstField = this.pageNode.querySelector(
+                "input:not(:disabled):not([readonly]):not([type=hidden])" +  
+                ",select:not(:disabled):not([readonly])"+
+                ",textarea:not(:disabled):not([readonly])");
+            firstField && firstField.focus();
+            this.pageNode && this.pageNode.parentNode.scrollIntoView({ behavior: "instant", block: "start", });
+        }
+    }
     render(): JSX.Element {
         if (this.page == null || this.survey == null || this.creator == null) return null;
         var title = this.renderTitle();
@@ -33,7 +54,7 @@ export class SurveyPage extends React.Component<any, any> {
             rows.push(this.createRow(questionRows[i], i));
         }
         return (
-            <div>
+            <div ref={node => this.pageNode = node }>
                 {title}
                 {rows}
                 </div>
--- a/src/react/reactquestion.tsx
+++ b/src/react/reactquestion.tsx
@@ -29,20 +29,43 @@ export class SurveyQuestion extends React.Component<any, any> {
         this.questionBase = question;
         this.question = question instanceof Question ? question : null;
         var value = this.question ? this.question.value : null;
-        this.state = { visible: this.questionBase.visible, value: value, error: 0, renderWidth: 0 };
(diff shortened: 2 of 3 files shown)


andrewtelnov:
Hi @jdalegonzalez, I have fixed the bug for knockout and react versions, could you please check it out?
I did it on the model level: https://github.com/andrewtelnov/surveyjs/commit/e63852cd878b23a8f86a20c5b12ed64387dbe122

Thank you,
Andrew

jdalegonzalez:
Will do. Thanks

jdalegonzalez:
It does correctly scroll to the top of the page but doesn't set focus to the first element.  Maybe just the focusing of the first element stuff should come over?

jdalegonzalez:
How about adding this code to reactpage.tsx (or somewhere else if that makes more sense)...
```javascript
import {SurveyPageId} from "../base";
```
```javascript
    componentDidMount() {
        this.focusFirstInput();
    }
    componentDidUpdate() {
        this.focusFirstInput();
    }
    focusFirstInput() {
        if (!this.page.isVisible || !this.page.isActive) return;
        var el = document.getElementById(SurveyPageId);
        if (!el || !el.querySelector) return;
        var firstInput =  el.querySelector(
            "input:not(:disabled):not([readonly]):not([type=hidden])" +  
            ",select:not(:disabled):not([readonly])"+
            ",textarea:not(:disabled):not([readonly])"
        );
        firstInput && (firstInput as HTMLElement).focus();
    }
```

andrewtelnov:
@jdalegonzalez Yes, it make sense. I will think, how to make this for both versions. If I do not come with any idea - I will accept your pull request.

jdalegonzalez:
For what it's worth, adding this code to reactquestion.tsx makes focus setting to the errored field work whereas before it didn't but it doesn't help with the initial focus setting on page show/hide

```javascript
    componentDidMount() {
        if (this.questionBase) {
            var self = this;
            this.questionBase["react"] = self;
            this.questionBase.renderWidthChangedCallback = function () {
                self.state.renderWidth = self.state.renderWidth + 1;
                self.setState(self.state);
            }
            this.questionBase.visibleIndexChangedCallback = function() {
                self.state.visibleIndexValue = self.questionBase.visibleIndex;
                self.setState(self.state);
            }
            this.questionBase.focusCallback = function() {
                var el = document.getElementById(self.questionBase.id);
                if (!el || !el.querySelector) return;
                var firstInput =  el.querySelector(
                    "input:not(:disabled):not([readonly]):not([type=hidden])" +  
                    ",select:not(:disabled):not([readonly])"+
                    ",textarea:not(:disabled):not([readonly])"
                );
                firstInput && (firstInput as HTMLElement).focus();
            }
        }
    }
    componentWillUnmount() {
        if (this.questionBase) {
            this.questionBase["react"] = null;
            this.questionBase.renderWidthChangedCallback = null;
            this.questionBase.visibleIndexChangedCallback = null;
            this.questionBase.focusCallback = null;
        }
    }
```

andrewtelnov:
@jdalegonzalez I want to make the next release and hope to do some experiments after that.
I did not quite understand your last message. I want to focus the fist input on the page on changing the current page. On the first survey showing, this code should not be executed - I believe on many websites it would not an expected behavior. However, there should be a method to call it manually.

jdalegonzalez:
Regarding my last message: There are three times during the life of a survey when you *might* want focus to be set to a field: at the beginning of the survey (but as you pointed out maybe not always), as you switch from one page to another and when there is an error in validation.  The code I offered in the last message lets the react code correctly set focus to a field if there has been an error in validation.  It doesn't solve the setting focus on a page transition or when the survey is initially displayed.

Thanks again for everything you're doing.

andrewtelnov:
@jdalegonzalez Could you please check the react version? https://github.com/andrewtelnov/surveyjs/issues/145

Thank you,
Andrew

jdalegonzalez:
Will do.

jdalegonzalez:
Looks like it works with two caveats.  

1. On an error, the focus is correctly set to input element that has the error but this means that the error message (which appears above the input element) may be scrolled off the screen.  The right thing *might* be to ensure that the error message div is visible and then set focus to the input element.

2. On a page transition, focus is set to the first element - which makes perfect sense when moving forward through the survey.  Does it make sense when moving backwards (previous) through the survey or should focus be set on the last element on the page?

As far as setting focus when the survey loads, is there a recommended process for doing this?


Thanks again!

andrewtelnov:
@jdalegonzalez I've just added a new method to the survey class (survey model): focusFirstQuestion() 
It scroll to the top of the current page and focus the the first input.

Thank you,
Andrew

PR #184, opened 2017-01-11
fixed bug https://github.com/andrewtelnov/surveyjs/issues/183

dmitrykurmanov:


andrewtelnov:
Solution that uses onMouseOut event could not be used :-(

PR #192, opened 2017-01-20
Edit & Feature addition

mellogarrett:
1) Fixed some weird wording to improve readability

2) FEATURE ADDITION: Added ability to submit partial submissions using your own custom logic. Just pass `sendResultOnPageNext: true` in Survey model, and specify your custom logic in `onPartialSend`.

Example usage: 

    $("#survey").Survey({
        model: { 
            ...
            sendResultOnPageNext: true
            ...
        },
        onComplete: sendFullResultToServer,
        onPartialSend: sendPartialResultToServer
    });


andrewtelnov:
@mellogarrett 
partialSend makes sense. I will modify it after merging. 'sendResult' method is not called when clientId is not set in doNextPage, that is not correct.

Thank you for the pull requests.

Andrew-

andrewtelnov:
It should be fine now.

Thank you,
Andrew

PR #261, opened 2017-02-25
add additional custom css entrypoints for dropdown and radiogroup

bennettrogers:
This adds a custom CSS entrypoint to the dropdown control and updates another on the radiogroup. This makes it easier to style the input elements with third-party CSS libraries (like [Bulma](http://bulma.io/)).

Currently the radiogroup control uses the same CSS class for both the div surrounding each option and the option label itself. This update changes that to use a separate CSS class for the label.

This also adds a new CSS class on the div surrounding the dropdown/select control to allow more detailed styling of that input element.

### DIFF
--- a/src/defaultCss/cssbootstrap.ts
+++ b/src/defaultCss/cssbootstrap.ts
@@ -14,12 +14,12 @@ export var defaultBootstrapCss = {
 
     checkbox: { root: "form-inline", item: "checkbox", other: "" },
     comment: "form-control",
-    dropdown: "form-control",
+    dropdown: { root: "", control: "form-control" },
     matrix: { root: "table" },
     matrixdropdown: { root: "table" },
     matrixdynamic: { root: "table", button: "button" },
     multipletext: { root: "table", itemTitle: "", itemValue: "form-control" },
-    radiogroup: { root: "form-inline", item: "radio", other: "" },
+    radiogroup: { root: "form-inline", item: "radio", label: "", other: "" },
     rating: { root: "btn-group", item: "btn btn-default" },
     text: "form-control",
     window: {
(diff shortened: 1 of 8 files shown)


andrewtelnov:
@bennettrogers Thank you for the pull request.

You are using dropdown.root only for rendering dropdown as div in "display" mode. Probably it is better to name it as display:  dropdown: { control: "", display: "" }.
What do you think?
I may do a correction by myself of course.

Thank you,
Andrew

bennettrogers:
First of all, thank you for an excellent library - I came across SurveyJS a while ago, and was so happy to see when the Vue version was released.

I think I see what you're saying regarding the naming here - my intention was to name the class at the top of the item "root", and name the class of the actual dropdown/select "control". I'm looking at my change now, and I realize what you mean. I'm just learning SueveyJS, so I wasn't familiar with display mode, but now I can see that in the React version, I made the toplevel classname "root" for display mode, and "control" for edit mode. 

I'm using the Vue version, so I was mainly trying to get that working, and I think it is done the way I intended - is that correct?

I will try to update my PR to make the naming consistent across the versions.

(Also, I've since noticed that there are a few other places that I'd like to add configurable classnames. Would you be comfortable with me submitting more pull requests for those, or would you prefer to add them yourself?)

bennettrogers:
Ok, I _think_ this commit will fix some of this. I couldn't figure out where the surrounding `<div>` in the knockout version should come from though, so right now I don't think it will use the dropdown.root class.

Just to make sure it's clear what I was trying to do, in all versions the end result should look something like:

Edit mode:
```
<div class={css.dropdown.root}>
  <select class={css.dropdown.control}>
    ...
  </select>
</div>
```

Display mode:
```
<div class={css.dropdown.root}>
  <div class={css.dropdown.control}>
    ...
  </div>
</div>
```

andrewtelnov:
@bennettrogers Yes, thank you!

PR #456, opened 2017-06-20
More accurate spellings for german localization.

ifcologne:


ifcologne:
Okay, the test requires "Fertig" as label for the submit button to complete a survey. 
"Fertig" is more like "done" - a little bit informal but not wrong. "Absenden" means the survey is completed and pushing the button "sends" the feedback to the server.

I would recommend "Absenden" but don't mind to stay with "Fertig". Everyone is used to surveys and will instinctively do the right next step.

dmitrykurmanov:
@ifcologne, I've fixed tests according to new german localization

dmitrykurmanov:
@ifcologne thanks for your help!

Issue #462, opened 2017-06-21
Numeric validator does not validate negative numbers if minValue: 0


dmitrykobets:
### Are you requesting a feature, reporting a bug or ask a question?

reporting a bug

### What is the current behavior?

Setting a numeric validator with a minValue: 0 does not give validation error for negative numbers.

### What is the expected behavior?

A numeric validator with a minValue: 0 should give validation error for any number < 0

### Specify your

* browser: Chrome
* browser version: 58.0.3029.110
* surveyjs platform (angular or react or jquery or knockout or vue): jquery
* surveyjs version: 0.12.17

dmitrykobets:
I have a fix for this so if you add me as a collaborator I could throw it in.

gruz:
At least in 0.12.23 the validation is still broken in this case.

Here is my form
```
{
 pages: [
  {
   elements: [
    {
     type: "text",
     inputType: "number",
     isRequired: true,
     name: "email",
     title: "Enter number",
     validators: [
      {
       type: "numeric",
       maxValue: 100,
       minValue: 1
      }
     ]
    }
   ]
  }
 ]
}
```

It allows to complete the form when entering zero.

And if I set minValue to zero in the form JSON, then the survey editor strips the parameter.
At my test site in this case it treats non-numbers as nothing, as if there is nothing entered. But it should show an error message.

So with minValue=1 it allows to enter zero.
With minValue=0 it allows to enter non-numbers.

andrewtelnov:
@gruz Yes, there was a bug. The unit test is added and fixed. The fix will be available in the next minor update, likely next week.

Thank you,
Andrew

tyzzzon:
Maybe I'm doing something wrong but number min/max validation doesn't work for me at all. I tried it in v 1.7.5 and v 1.8.4. I tried it in survey creator.
My element is 
```
{
    "type": "text",
    "name": "1",
    "title": "Number of times per week =",
    "titleLocation": "left",
    "min": 0,
    "max": 100,
    "requiredErrorText": "Please enter a value of 0 or greater",
    "inputType": "number"
}
```
It accepts any number. But it doesn't allow you to enter letters(as expected)

andrewtelnov:
@tyzzzon Please put min and max bellow inputType and it will work.
I will investigate the issue.

Thank you,
Andrew

andrewtelnov:
[...]

PR #464, opened 2017-06-22
Fixed bug in issue #462. Added corresponding tests

dmitrykobets:


andrewtelnov:
@dmitrykobets The fix looks fine to me. For some reason some function tests do not passed. I have restarted them.

Thank you,
Andrew

andrewtelnov:
@dmitrykobets Functional tests do not work because the following service: http://services.groupkt.com/country/get/all is down.
We will re-write those functional tests.

andrewtelnov:
Thank you for the fix!

PR #506, opened 2017-07-19
added support for simplified Chinese

xraywu:
Have added support for simplified Chinese.

BTW. Where can we add simplified Chinese to the survey locale property options in the editor?


xraywu:
Does the build failure have anything to do with my codes? Although does not seem to...

andrewtelnov:
@xraywu I have confirmed the pull request. The issue in our functional tests. They are not 100% stable yet. We will take a look at them.

Thank you,
Andrew

PR #617, opened 2017-09-01
Extract class active in rating

balping:
This PR adds a possibility to customise the class applied on a selected rating button. This might be different in other css frameworks than bootstrap (eg. `is-active` in foundation and bulma)


andrewtelnov:
Yes, it makes sense.

Thank you! 
Andrew

PR #642, opened 2017-09-09
Add Persian Translation

cracki:


andrewtelnov:
@cracki Thank you!

Andrew

PR #730, opened 2017-10-24
Update React typings version in package.json

ahtsei:
This update is required if working with Typescript version >= 2.4.0 - related to this it seems: https://gist.github.com/DanielRosenwasser/f5e7cbee0241b8edc1830233b160b46c#weak-types

The proposed React types version update here fixes this.  I came across this issue by trying to use string Enums in Typescript, which got released in 2.4.0. Should avoid headaches in the future.


ahtsei:
Thanks for that fix @andrewtelnov  👍

PR #749, opened 2017-11-03
Added ^ to React dependency version 

ahtsei:
Have updated the webpack config file to use a newer version of React (same as the dev dependency in package.json).


tsv2013:
We'd like to guarantee that our library works with certain version of the dependency.
We locked dependencies to avoid suddenly compilation breaks caused by breaking changes, even unexpected.

Issue #818, opened 2017-12-11
Survey-Vue: page.vue and panel.vue missing `hasDescription` computed property


benirose:
### Are you requesting a feature, reporting a bug or ask a question?
Reporting a Bug.

### What is the current behavior?
Getting Javascript Error. 
```
[Vue warn]: Property or method "hasDescription" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <SurveyPage> at /home/travis/build/surveyjs/surveyjs/src/vue/page.vue
       <Survey> at /home/travis/build/surveyjs/surveyjs/src/vue/survey.vue
         <App> at src/App.vue
           <Root>
```

Looks like this computed property was added to the template of `page.vue` in 3f37847, but it was never defined. I believe `panel.vue` has the same issue.

### What is the expected behavior?
Looks like `hasDescription` is supposed to be a computed property like `hasTitle`, but it's not defined. 

### How would you reproduce the current behavior (if this is a bug)?
Using SurveyJS Vue library, instantiate a `<survey>` component.

#### Provide the test code and the tested page URL (if applicable)

I tried to recreate this bug in a Codepen using [this CDN delivered version](https://cdnjs.cloudflare.com/ajax/libs/survey-vue/0.98.1/survey.vue.js), but it doesn't seem to have this bug yet, even though the version number is the same as what I get through NPM. When checking out the [Quickstart](https://github.com/surveyjs/surveyjs_vue_quickstart) and changing package.json to use the latest version of `survey-vue`, I was able to reproduce the error.

### Specify your

* browser: Chrome
* browser version: 62
* surveyjs platform (angular or react or jquery or knockout or vue): Vue
* surveyjs version: 0.98.1

PR #819, opened 2017-12-11
Implement computed property for hasDescription in Page/Panel for vue: #818

benirose:
I believe this should fix issue #818, but I wasn't 100% sure if it's the correct business logic. I dug around and didn't see a `showDescription` property like there is with title, so I just implemented this getter to return true/false based on description alone.

### DIFF
--- a/src/vue/page.vue
+++ b/src/vue/page.vue
@@ -43,6 +43,9 @@
         get hasTitle () {
             return !!this.page.title && this.survey.showPageTitles;
         }
+        get hasDescription () {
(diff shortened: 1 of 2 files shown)


benirose:
It does not look like the failing build is related to my changes.

andrewtelnov:
@benirose Yes, there is a functional test that has to be changed. 
You fix is completely fine.

andrewtelnov:
@benirose Of course it is there. We have released v0.98.3 today :)

Thank you,
Andrew

Issue #932, opened 2018-02-15
Please supply a :key to all v-for in Vue components


kaphert:
### Are you requesting a feature, reporting a bug or ask a question?
Bug

Related to https://github.com/surveyjs/surveyjs/issues/930

`v-for` for VueJS does not requires a `:key` attribute but without it is has strange behavior when the dom is changed or reordered. See https://vuejs.org/v2/guide/list.html#key

> It is recommended to provide a key with v-for whenever possible, unless the iterated DOM content is simple, or you are intentionally relying on the default behavior for performance gains.

Take example: https://plnkr.co/edit/GCKFIVh3PtVV3st99nKo?p=preview.
A custom widget is used here inside a dynamic panel. The widget it self is pretty stupid, it just creates a dropdown with 4 values. But it has some strange behavior:

1. Press 2 times on 'Add New'. We now have 3 panels, for simplicity we call them A, B and C.
2. Select a value for panel B (second)
3. Remove panel B
4. Notice that panel C now shows the value that we had selected on panel B? 
5. Now press on 'Complete'. Notice that it showing a 'please answer the question' error because he doesnt have a value. 

All this behavior in combination with #930 smells like the 'v-for behavior without a :key parameter`. 

Just confirmed that the same setup with jQuery does not result in this behavior: https://plnkr.co/edit/nKeYYsv9AvTlP0Qej3Kr?p=preview

tsv2013:
It's very strange that vue doesn't complain the "key" is mandatory in fact.

tsv2013:
I've added `:key` attribute, but this doesn't help with #930

kaphert:
@tsv2013 thanks but the `:key` values are not unique keys because the `inputId` property is undefined. If I fix this locally it solves the issue of #930 & the issue I described above. If there is a unique id for the element that you are looping at the moment that will be enough. So just passing `question.id`, `page.id`, `panel.id` to the `:key` will be enough. 

A row does not have a unique id so a combination of the page id or panel id and the row index would be good: `:key="page.id +'_'+ index". 

For all the checkbox, option, and radiobuttons the key could be equal to the value of the `item`. I think you can even leave out the :key here but better safe than sorry.

kaphert:
@tsv2013 do you want me to make a pull request to change the `:key` values to unique values?

dmitrykurmanov:
No problem. Sure.

dmitrykurmanov:
[...]

Issue #930, opened 2018-02-15
[Bug] onAfterRenderQuestion has unexpected behavior on survey with pages


kaphert:
### Are you requesting a feature, reporting a bug or ask a question?
Bug

### What is the current behavior?
I have 2 pages, first page has 2 questions, second page 1. The onAfterRenderQuestion is only called twice for the first 2 questions on the first page. 

### What is the expected behavior?
It should be called for each question

### How would you reproduce the current behavior (if this is a bug)?

1. First page 2 questions, second page 1 question. onAfterRenderQuestion is only called for the first two questions on page 1. see https://plnkr.co/edit/nQIrcOvbXwvJyejCSt8x?p=preview
2. First page has 1 question, second page has 2 questions. onAfterRenderQuestion is only called for the first question on page 1, and second question of page 2. see https://plnkr.co/edit/Fuei3y9bMuVTGfyyE6DB?p=preview

### Specify your

* surveyjs platform (angular or react or jquery or knockout or vue): Vue
* surveyjs version: v1.0.8

tsv2013:
I've added the `:key` attribute, but this doesn't fix this issue. It also doesn't fix dynamic panel issue from the sample attached to #932.

This question certainly is not related to the `:key` attribute presence.

This behaviour is exaplined pretty simple.
Please take a look at this code:
https://github.com/surveyjs/surveyjs/blob/35f8ee5c1178b09b46683714ff1030faabe34fd4/src/vue/element.vue#L50-L54

The `afterRenderQuestion` event is fired then component has been mounted - i.e. really rendered into the DOM according to the VueJS lifecycle diagram - https://vuejs.org/v2/guide/instance.html#Lifecycle-Diagram

That's why `afterRenderQuestion` event fires on the first page 2 times (2 components mount to the DOM), not fires on the second page (first component is being updated, second - destroyed), and fires once on return to the first page (one component reused and updated, one component - mounted).

This looks correct for me, because if you create some heavy custom control (e.g. Ace text editor) during this event, you will not need to re-create it on every update - you should just subscribe to the current question change and update the vallue assigned to the control.

We will be glad to get recommendations from VueJS professionals - what will be the best solution for this case.

kaphert:
This ain't correct behavior because this defeats the purpose of the afterRenderQuestion() functionality. And the JQuery variant of the same code has different behavior. 

[...]

PR #941, opened 2018-02-19
Work for #932 - make :key for v-for as unique as possible

kaphert:
Makes the :key unique for each v-for. This is v-for that contain other components. This way vue can better track if the component must be re-rendered when the collection for the v-for changes. 

This commit will solve #930

### DIFF
--- a/src/vue/file.vue
+++ b/src/vue/file.vue
@@ -4,7 +4,7 @@
         <button v-if="hasValue" :class="question.cssClasses.removeButton" @click="doClean">{{question.cleanButtonCaption}}</button>
         <input v-if="question.isReadOnly" type="text" readonly :class="getPlaceholderClass()" :placeholder="question.title" />
         <div v-if="hasValue">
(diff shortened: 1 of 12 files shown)


src/vue/file.vue
```diff
@@ -4,7 +4,7 @@
         <button v-if="hasValue" :class="question.cssClasses.removeButton" @click="doClean">{{question.cleanButtonCaption}}</button>
         <input v-if="question.isReadOnly" type="text" readonly :class="getPlaceholderClass()" :placeholder="question.title" />
         <div v-if="hasValue">
-            <img v-for="(val, index) in question.previewValue" :key="question.inputId + 'key' + index" v-show="val" :class="question.cssClasses.preview" :src="val" :height="question.imageHeight" :width="question.imageWidth" alt="File preview"/>
+            <img v-for="val in question.previewValue" :key="val" v-show="val" :class="question.cssClasses.preview" :src="val" :height="question.imageHeight" :width="question.imageWidth" alt="File preview"/>
```
tsv2013: 'val' here is a base64-encoded file content - it shouldn't be used as a key

tsv2013:
I've reverted "file.vue" to original due to the "val" is too long for key. Other changes looks ok for me. After tests passed I'll mergre it.
Thank you, @kaphert

PR #942, opened 2018-02-20
Add has-error class to element component for Vue

kaphert:
This commit adds the class defined in the `question.hasError` to the element in the html. This allows us to easily customize the look for the form fields and the titles when an error pops up. 

Note: this commit only adds it to the vue components. If the idea behind this is ok it should be added to the other frameworks also.

### DIFF
--- a/src/defaultCss/cssbootstrap.ts
+++ b/src/defaultCss/cssbootstrap.ts
@@ -30,6 +30,7 @@ export var defaultBootstrapCss = {
     comment: "form-control",
     required: "",
     titleRequired: "",
+    hasError: "has-error",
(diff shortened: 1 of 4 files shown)


tsv2013:
Whether the 'has-error' class should be added for other platforms (React, Knockout)?

kaphert:
Yes indeed. In this commit it is only added to the Vue platform templates.

tsv2013:
See also [Added "has error" class for knockout](https://github.com/surveyjs/surveyjs/commit/75ba24223bd9655751ac8329f270a295014e5ffa)

PR #1422, opened 2018-11-02
Fix #1417

### LINKED ISSUE #1417: Sometimes first question on a page does not auto-focus

Bug

Sometimes, after clicking to go to the next survey page, the first question is not auto-focused.

I think this happens because the survey tries to determine which question to auto-focus by finding the first question object on the page with `visible=true`, but does not check if its parent object (page/panel) is `visible=true` as well.

So if a user set the `visibleIf` property for a **panel** in the survey JSON, even though its questions may technically be "hidden", the questions still have `visible=true` by default, and the survey tries to focus on it.

The first question should be auto-focused.

1) Go to any survey that shows different pages/panels depending on previous answers.
(I'm sorry in advance, since I don't have a sample URL on hand, and the survey I'm developing needs an account to access it).

2) When reaching a question where, for example, Option A displays Panel A on the next page, and Option B displays Panel B, choose Option B (the second or later options, not the first).

3) Notice that first question on next page is not auto-focused.

#### Provide the test code and the tested page URL (if applicable)

Tested page URL: none

Sample Survey JSON that would produce this issue

```
{
	"pages": [{
		"name": "page1",
		"elements": [{
			"type": "radiogroup",
			"name": "QUESTION1",
			"title": "Which unit of measurement do you prefer?",
			"isRequired": true,
			"choices": [{
					"value": "M",
					"text": "Metric"
				},
				{
					"value": "I",
					"text": "Imperial"
				}
			]
		}]
	}, {
		"name": "page2",
		"elements": [{
				"type": "panel",
				"name": "QUESTION2_CM",
				"elements": [{
					"type": "text",
					"name": "QUESTION2_CM_Q",
					"title": "Centimeters (cm)",
					"isRequired": true,
					"validators": [{
						"type": "numeric",
						"minValue": 60,
						"maxValue": 240
					}],
					"placeHolder": "Enter a number"
				}],
				"visibleIf": "{QUESTION1} = \"M\"",
				"title": "How tall are you?"
			},
			{
				"type": "panel",
				"name": "QUESTION2_FT",
				"elements": [{
						"type": "text",
						"name": "QUESTION2_FT_Q",
						"title": "Feet",
						"isRequired": true,
						"validators": [{
							"type": "numeric",
							"minValue": 2,
							"maxValue": 8
						}],
						"placeHolder": "Enter a number"
					},
					{
						"type": "text",
						"name": "CC_06_I_inches",
						"startWithNewLine": false,
						"title": "Inches",
						"isRequired": true,
						"validators": [{
							"type": "numeric",
							"minValue": 0,
							"maxValue": 11
						}],
						"placeHolder": "Enter a number"
					}
				],
				"visibleIf": "{QUESTION1} = \"I\"",
				"title": "How tall are you?"
			}
		]
	}]
}

```
If on Page 1, the 2nd option "Imperial" was chosen, when going to Page 2, the question is not focused.

### Specify your

* browser: Chrome
* browser version: 69.0.3497.100
* surveyjs platform (angular or react or jquery or knockout or vue): react
* surveyjs version: 1.0.53

ekim-novus:
When giving focus to the first question on the page, check if question's parent (page/panel) is visible as well (to avoid giving focus to a hidden question)

### DIFF
--- a/src/page.ts
+++ b/src/page.ts
@@ -134,7 +134,12 @@ export class PageModel extends PanelModelBase implements IPage {
   public focusFirstQuestion() {
     for (var i = 0; i < this.questions.length; i++) {
       var question = this.questions[i];
-      if (!question.visible || !question.hasInput) continue;
+      if (
+        !this.questions[i].parent.isVisible ||
+        !question.visible ||
+        !question.hasInput
+      )
+        continue;
       this.questions[i].focus();
       break;
     }
@@ -145,6 +150,7 @@ export class PageModel extends PanelModelBase implements IPage {
   public focusFirstErrorQuestion() {
     for (var i = 0; i < this.questions.length; i++) {
       if (
+        !this.questions[i].parent.isVisible ||
         !this.questions[i].visible ||
         this.questions[i].currentErrorCount == 0
       )


andrewtelnov:
Unfortunately, the PR doesn't fix the issue completely. I have provided an alternative fix.

Thank you,
Andrew

PR #1549, opened 2019-02-14
build: use minified file as main on production builds

vcanales:
With this change, when creating production builds, the `package.json` file will load the minified version instead of the full version. This saves **~2.4mb** on production, which is critical to page load times.

I did not update the vue build, since it appears to be broken regardless.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -149,7 +149,7 @@ module.exports = function(options) {
   var mainFile =
     options.platform === "vue"
       ? "survey-vue.js"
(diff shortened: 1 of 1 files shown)


dmitrykurmanov:
Hello! Thank you for the contribution. But we are not sure what is the best way to deliver commonjs module. Please take a look to https://github.com/surveyjs/surveyjs/issues/495 we changed "min" version to "unmin" via this issue. It would be great If you would share with us some articles about best practice  npm umd module delivering.

PR #1565, opened 2019-02-28
Added japanese translation

Tammy05:


andrewtelnov:
@Tammy05 Thank you!
Andrew

PR #1799, opened 2019-08-19
Add click handler to cell in matrix questions

jeffcann:
Currently for cells in matrix questions a user has to click/touch on the radio button.
In mobile mode we have styled the cells to look like buttons but they only register a click on the radio button or label. A click handler on the table cell gives us the exact behaviour we need.


tsv2013:
@jeffcann Thank you for the contribution!

PR #1891, opened 2019-11-05
add authorization to choicesByUrl

Evertude:
I'd like to add a bearer token to my rest service call because currently I have to override the onBeforeSendRequest method in order to set the header


tsv2013:
This can be achieved in the "Survey.ChoicesRestfull.onBeforeSendRequest" callback

Here are the dicsussions and sample code links
https://surveyjs.answerdesk.io/ticket/details/T405
https://surveyjs.answerdesk.io/ticket/details/T2439

Evertude:
I'm aware of this possibility and I'm also currently using it but I see two issues:
- question_baseselect has a default behaviour which gets overriden
- onBeforeSendRequest is static which means that if I'd like to use two different endpoints that require different tokens, I'd have to remember which one to use for what

PR #1988, opened 2020-02-07
Updated italian localization

msamuelsson:
Added translation for File related messages.


tsv2013:
Thank you for the contribution!

PR #2054, opened 2020-03-27
Added missing localization string for german language

bludau-peter:


andrewtelnov:
@bludau-peter Thank you!

Andrew

PR #2072, opened 2020-04-08
master

vfaidherbe:
Added missing translation for dutch localization files


andrewtelnov:
@vfaidherbe Great and thank you for the second language!

Andrew

PR #2194, opened 2020-05-27
Tajik locale

vladshlyanin:
Created tajik localization file


andrewtelnov:
@vladshlyanin 
Thank you!
Andrew

PR #2281, opened 2020-07-14
Fix: header attribute value in matrixdynamic cells

feedmypixel:
# Populate matrixdynamic table cell header attribute with column title value

I have just updated `surveyjs` from `v^1.0.73` to `v^1.7.19` and we have an issue:

The header attribute value for a `matrixdynamic` table cell was not being populated with a columns title value. This meant that when a survey is viewed in small screen there is no label for an input.

This fix locates the `this.cell.cell.column.locTitle.renderedHtml` on the cells column within the `getHeaderText` method and returns it to be applied to the table cells `header` attribute.

#### Large screen
Screenshot 2020-07-14 at 14 53 59

#### Small screen
Screenshot 2020-07-14 at 14 54 13

#### Large screen
Screenshot 2020-07-14 at 14 54 44

#### Small screen
Screenshot 2020-07-14 at 14 54 33

## Tests
- I have ran all tests and we are green
- I have also built the project locally and used it in our app. Everything works as expected
- I have looked around to add a test for this but cannot see any react tests around

### New to surveyjs
I am new to `surveyjs` and would appreciate any thoughts on this fix as I may of missed something. I am certainly seeing `this.cell.cell` around the codebase so feel this is an approach taken elsewhere.

### Example have this problem
You can also observe that the table cells `header` attribute is not being populated on [surveyjs matrixdynamic example](https://surveyjs.io/Examples/Library?id=questiontype-matrixdynamic&platform=Reactjs&theme=modern#content-result)

Screenshot 2020-07-14 at 15 08 08

### DIFF
--- a/src/react/reactquestionmatrixdropdownbase.tsx
+++ b/src/react/reactquestionmatrixdropdownbase.tsx
@@ -229,7 +229,8 @@ export class SurveyQuestionMatrixDropdownCell extends SurveyQuestionAndErrorsCel
   }
 
   protected getHeaderText(): string {
-    return !!this.cell.locTitle ? this.cell.locTitle.renderedHtml : "";
+    var column = this.cell.cell && this.cell.cell.column;
+    return !!(column && column.locTitle) ? column.locTitle.renderedHtml : "";
   }
   protected renderQuestion(): JSX.Element {
     if (!this.cell.isChoice)


tsv2013:
Thank you for the contribution!

feedmypixel:
@tsv2013 when are you planning to release [v1.7.20](https://github.com/surveyjs/survey-library/milestone/29)? You look close. Could you **@** me in the release please?

tsv2013:
In the nearest couple of days

tsv2013:
You can subrcribe this repo to get notified

PR #2315, opened 2020-08-05
Updated english.ts

MrKrabat:
Hello,

In my opinion, urlRequestError/urlGetChoicesError and cleanCaption/clearCaption in particular needed a change. The error messages should also be understandable for an end user.

What's your opinion?


src/localization/english.ts
```diff
@@ -18,10 +18,10 @@ export var englishStrings = {
   completingSurvey: "Thank you for completing the survey!",
   completingSurveyBefore:
     "Our records show that you have already completed this survey.",
-  loadingSurvey: "Loading Survey...",
+  loadingSurvey: "Loading survey...",
   optionsCaption: "Choose...",
   value: "value",
-  requiredError: "Please answer the question.",
+  requiredError: "Please answer this question.",
```
dirk-pieterse, line 24: I'm not sure why 'this' instead of 'the'. I would keep 'the'.

src/localization/english.ts
```diff
@@ -37,14 +37,14 @@ export var englishStrings = {
   numericMax: "The '{0}' should be equal or less than {1}",
   invalidEmail: "Please enter a valid e-mail address.",
   invalidExpression: "The expression: {0} should return 'true'.",
-  urlRequestError: "The request returned error '{0}'. {1}",
+  urlRequestError: "A network service has returned the following error '{0}'. {1}",
```
dirk-pieterse, line 40: - 'A network service' is quite vague. Which network service'? What causes the error?
- I would change 'has returned' to 'returned' since you are referring to the past tense.

src/localization/english.ts
```diff
@@ -37,14 +37,14 @@ export var englishStrings = {
...
   urlGetChoicesError:
-    "The request returned empty data or the 'path' property is incorrect",
+    "A network service has returned invalid data",
```
dirk-pieterse, line 42: Same here

src/localization/english.ts
```diff
@@ -18,10 +18,10 @@ export var englishStrings = {
...
   completingSurveyBefore:
     "Our records show that you have already completed this survey.",
-  loadingSurvey: "Loading Survey...",
+  loadingSurvey: "Loading survey...",
   optionsCaption: "Choose...",
   value: "value",
-  requiredError: "Please answer the question.",
+  requiredError: "Please answer this question.",
   requiredErrorInPanel: "Please answer at least one question.",
   requiredInAllRowsError: "Please answer questions in all rows.",
```
dirk-pieterse, line 26: 'Please answer questions in all rows' What do you mean by all rows? If every row has a question, you could just say 'Please answer all the questions'.

dirk-pieterse (approved):

src/localization/english.ts
MrKrabat: I'm not a native English speaker, so maybe I write nonsense.
This error message refers to a specific question that needs to be answered. We want the user to answer 'this' question. So I think 'this' is grammatically better.

src/localization/english.ts
MrKrabat: I did not change this string, but you are right. It should be changed.

src/localization/english.ts
MrKrabat: This error is displayed to users who probably have no knowledge of surveyjs, computer and network stuff. An error message that appears with "the request returned error ..." does not tell the user what is wrong. You are right, there is no context that loading answers from external resources failed.

This message should give the user an indication of what is going wrong so that the user can send a helpful bug report to the survey creator.

PR #2441, opened 2020-10-07
Change "Previous" translation for Japanese

hoechenberger:
We have deployed a survey in Japanese, and received the feedback that the current Japanese default translation of the "Previous" button is incorrect or at least unusual. This PR fixes this and uses the correct term (which was confirmed to us as being correct through several people whose first language is Japanese).

### DIFF
--- a/src/localization/japanese.ts
+++ b/src/localization/japanese.ts
@@ -2,7 +2,7 @@
 import { surveyLocalization } from "../surveyStrings";
 
 export var japaneseSurveyStrings = {
-  pagePrevText: "前の",
+  pagePrevText: "前へ",
   pageNextText: "次へ",
   completeText: "完了",
   startSurveyText: "スタート",


andrewtelnov:
@hoechenberger Are you sure? I did not speak Japanese at all. However [here is what google translation give me](https://translate.google.com/#view=home&op=translate&sl=auto&tl=en&text=%E5%89%8D%E3%81%AE%0A%E5%89%8D%E3%81%B8).
The current translation looks better to me :-). Of course it could be absolutely wrong, but it would love to re-check it one more time.

Thank you,
Andrew

hoechenberger:
> @hoechenberger Are you sure? I did not speak Japanese at all. However [here is what google translation give me](https://translate.google.com/#view=home&op=translate&sl=auto&tl=en&text=%E5%89%8D%E3%81%AE%0A%E5%89%8D%E3%81%B8).
> The current translation looks better to me :-). Of course it could be absolutely wrong, but it would love to re-check it one more time.

Haha yes, I also used Google Translate and was irritated. I will ask another independent Japanese native speaker tomorrow and report back.

hoechenberger:
@andrewtelnov I just received feedback from a Japanese friend of mine. She said that the string in `master` is grammatically incorrect, and the proposal in this PR would be the correct phrasing.

So this should be good to merge 👍

PR #2444, opened 2020-10-08
Added missing translations

MrKrabat:
Since most translations were not even half complete, I added the missing translations.

What I did:
- I have changed the order of the translations to match the order of the english translation file.
- I added missing translations using DeepL and Google Translator
- I did **not** change existing translations

In my opinion machine translations are better than missing translations.


andrewtelnov:
@MrKrabat Looks good for languages I know. I really hope the others good too.
Thank you for great contribution!

Andrew

Issue #2538, opened 2020-11-25
React SurveyQuestionMultipleText Missing Unique "key" prop


hedgecox:
### Are you requesting a feature, reporting a bug or asking a question?

Bug

### What is the current behavior?

When running react in dev mode you see the following error:

```
Warning: Each child in an array or iterator should have a unique "key" prop.

Check the render method of `SurveyQuestionMultipleText`. See https://fb.me/react-warning-keys for more information.
    in span (created by SurveyQuestionMultipleText)
    in SurveyQuestionMultipleText (created by SurveyQuestion)
    in div (created by SurveyQuestion)
    in div (created by SurveyQuestion)
    in SurveyQuestion (created by SurveyRow)
    in div (created by SurveyRow)
    in SurveyRow (created by SurveyPage)
    in div (created by SurveyPage)
    in SurveyPage (created by Survey)
    in div (created by Survey)
    in div (created by Survey)
    in form (created by Survey)
    in div (created by Survey)
    in Survey
```

### What is the expected behavior?

Warning should not appear

### How would you reproduce the current behavior (if this is a bug)?

Behavior is visible in default react example page

#### Provide the test code and the tested page URL (if applicable)

Tested page URL: https://surveyjs.io/Examples/Library/?id=questiontype-multipletext&platform=Reactjs

```
var json = {
    questions: [
        {
            type: "multipletext",
            name: "pricelimit",
            title: "What is the... ",
            colCount: 2,
            items: [
                {
                    name: "mostamount",
                    title: "Most amount you would every pay for a product like ours"
                }, {
                    name: "leastamount",
                    title: "The least amount you would feel comfortable paying"
                }
            ]
        }
    ]
};
```

### Specify your

* browser: All
* surveyjs platform (angular or react or jquery or knockout or vue): React
* surveyjs version: 1.8.14

PR #2539, opened 2020-11-25
#2538 Add key prop to SurveyQuestionMultipleText

hedgecox:
#2538 Addition of key props for all spans that are added to the element to remove react error


andrewtelnov:
@hedgecox Thank you for the fix!

Andrew

PR #2628, opened 2021-01-25
Better Finnish translations

vyznev:
It's nice to see that there *is* a Finnish localization file, but some of the translations were kind of awkward or just wrong. These should be better.


src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
-  maxSelectError: "Ole hyvä ja valitse enintään {0} vaihtoehtoa.",
+  minSelectError: "Valitse vähintään {0} vaihtoehtoa.",
+  maxSelectError: "Valitse enintään {0} vaihtoehtoa.",
   numericMinMax:
-    "'{0}' täytyy olla enemmän tai yhtä suuri kuin {1} ja vähemmän tai yhtä suuri kuin {2}",
-  numericMin: "'{0}' täytyy olla enemmän tai yhtä suuri kuin {1}",
-  numericMax: "'{0}' täytyy olla vähemmän tai yhtä suuri kuin {1}",
+    "Luvun '{0}' tulee olla vähintään {1} ja korkeintaan {2}.",
+  numericMin: "Luvun '{0}' tulee olla vähintään {1}.",
+  numericMax: "Luvun '{0}' tulee olla korkeintaan {1}.",
```
vyznev, line 37: These weren't _wrong_ as such, but kind of klunky and verbose. (Honestly, so are the English strings too.) Prepending the word _luku_ ("number") to the field name allows correct noun conjugation (_luku_ &rarr; _luvun_), improving the grammar.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   invalidEmail: "Syötä validi sähköpostiosoite.",
   invalidExpression: "Lausekkeen: {0} pitäisi palauttaa 'true'.",
   urlRequestError: "Pyyntö palautti virheen {0}. {1}",
   urlGetChoicesError:
-    "Pyyntö palautti tyhjät tiedot tai polku-ominaisuus on väärä",
+    "Pyyntö palautti tyhjän tiedoston tai 'path'-asetus on väärä",
   exceedMaxSize: "Tiedoston koko ei saa olla suurempi kuin {0}.",
-  otherRequiredError: "Anna toinen arvo.",
+  otherRequiredError: "Tarkenna vastaustasi tekstikenttään.",
```
vyznev, line 44: The old string (which literally translated as "Give another value.") doesn't work at all in the context where it's used. The new translation literally means "Specify your answer in the text field." which is a lot clearer in context.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   uploadingFile:
-    "Tiedostoa ladataan. Odota muutama sekunti ja yritä uudelleen.",
+    "Tiedostoa lähetetään. Odota muutama sekunti ja yritä uudelleen.",
```
vyznev, line 46: Replaced _ladata_ ("(down)load") with _lähettää_ ("send, upload") to correctly describe the direction of the transfer.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   loadingFile: "Ladataan...",
-  chooseFile: "Valitse tiedosto (t) ...",
+  chooseFile: "Valitse tiedosto(t)...",
   noFileChosen: "Ei tiedostoa valittuna",
-  confirmDelete: "Haluatko poistaa tietueen?",
-  keyDuplicationError: "Tämän arvon tulisi olla ainutlaatuinen.",
+  confirmDelete: "Haluatko poistaa osion?",
+  keyDuplicationError: "Tämä arvo on jo käytössä. Syötä toinen arvo.",
```
vyznev, line 51: _Ainutlaatuinen_ means "unique" in the specific sense of "special, exceptional, only one of its kind", which is wrong here. Alas, there's no good Finnish word for "unique" in the sense of "not duplicate", so I ended up rephrasing this message to say, literally, "This value is already in use. Please choose another value."

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   removePanel: "Poista",
   choices_Item: "kohde",
   matrix_column: "Sarake",
   matrix_row: "Rivi",
-  savingData: "Tulokset tallennetaan palvelimelle ...",
-  savingDataError: "Tapahtui virhe, emmekä voineet tallentaa tuloksia.",
-  savingDataSuccess: "Tulokset tallennettiin onnistuneesti!",
+  savingData: "Tietoja tallennetaan palvelimelle...",
+  savingDataError: "Tapahtui virhe, emmekä voineet tallentaa kyselyn tietoja.",
+  savingDataSuccess: "Tiedot tallennettiin onnistuneesti!",
```
vyznev, line 63: _Tulokset_ literally means "results", which seems slightly off here. I replaced it with _tiedot_, which literally means "data" or "(items of) information" and feels more appropriate to me in general.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   saveAgainButton: "Yritä uudelleen",
   timerMin: "min",
   timerSec: "sek",
   timerSpentAll: "Olet käyttänyt {0} tällä sivulla ja yhteensä {1}.",
-  timerSpentPage: "Olet viettänyt {0} tällä sivulla.",
+  timerSpentPage: "Olet käyttänyt {0} tällä sivulla.",
```
vyznev, line 68: Used the verb _käyttää_ ("use") instead of _viettää_ ("spend (time)") for consistency with other messages here. Both would be valid, but _käyttää_ feels somewhat more active, with _viettää_ having connotations of idling away the time.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   timerSpentSurvey: "Olet käyttänyt yhteensä {0}.",
   timerLimitAll:
     "Olet käyttänyt tällä sivulla {0} / {1} ja yhteensä {2} / {3}.",
   timerLimitPage: "Olet käyttänyt {0} / {1} tällä sivulla.",
   timerLimitSurvey: "Olet käyttänyt yhteensä {0} / {1}.",
-  cleanCaption: "Puhdas",
-  clearCaption: "Asia selvä",
+  cleanCaption: "Pyyhi",
+  clearCaption: "Tyhjennä",
```
vyznev, line 75: These were both just plain wrong before: _puhdas_ is an adjective, not a verb, and "_asia selvä_" is an idiomatic phrase that means "clear" in the sense of "understood" or "OK, got it!".

I'm not sure where `cleanCaption` is used in the UI, so the translation _pyyhi_ (literally "wipe, erase") might not be optimal. Some other translations seem to use the same text for both of these strings, which (i.e. making both of them _tyhjennä_ = "clear, empty, drain") might be an option here for Finnish too.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   chooseFileCaption: "Valitse tiedosto",
   removeFileCaption: "Poista tämä tiedosto",
-  booleanCheckedLabel: "Joo",
+  booleanCheckedLabel: "Kyllä",
```
vyznev, line 78: _Kyllä_ is more formal than _joo_, which is basically equivalent to "yeah" or "yep" in English.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   confirmRemoveFile: "Haluatko varmasti poistaa tämän tiedoston: {0}?",
   confirmRemoveAllFiles: "Haluatko varmasti poistaa kaikki tiedostot?",
   questionTitlePatternText: "Kysymyksen otsikko",
+  modalCancelButtonText: "Peruuta",
+  modalApplyButtonText: "Käytä",
 };
 
 surveyLocalization.locales["fi"] = finnishSurveyStrings;
-surveyLocalization.localeNames["fi"] = "suomalainen";
+surveyLocalization.localeNames["fi"] = "suomi";
```
vyznev, line 88: _Suomalainen_ is an adjective describing something Finnish. The name of the language is simply _suomi_.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
   chooseFileCaption: "Valitse tiedosto",
   removeFileCaption: "Poista tämä tiedosto",
-  booleanCheckedLabel: "Joo",
+  booleanCheckedLabel: "Kyllä",
   booleanUncheckedLabel: "Ei",
   confirmRemoveFile: "Haluatko varmasti poistaa tämän tiedoston: {0}?",
   confirmRemoveAllFiles: "Haluatko varmasti poistaa kaikki tiedostot?",
   questionTitlePatternText: "Kysymyksen otsikko",
+  modalCancelButtonText: "Peruuta",
+  modalApplyButtonText: "Käytä",
```
vyznev, line 84: There's no exact idiomatic translation for "apply" that would work as a dialog button in Finnish, but _käytä_ ("use") comes pretty close, and is probably the standard choice for such dialogs. Other possible options could be _hyväksy_ ("approve, accept") or simply _OK_.

src/localization/finnish.ts
```diff
@@ -5,16 +5,16 @@ export var finnishSurveyStrings = {
   pageNextText: "Seuraava",
   completeText: "Valmis",
   previewText: "Esikatselu",
-  editText: "Muokata",
-  startSurveyText: "alkaa",
-  otherItemText: "Muu (kuvaile)",
+  editText: "Muokkaa",
+  startSurveyText: "Aloita",
+  otherItemText: "Muu (tarkenna)",
```
vyznev, line 10: _Tarkenna_ ("specify") seems more appropriate than _kuvaile_ ("describe, illustrate, characterize") here. The latter verb is something one would use in contexts like "Describe yourself as a team member" where one is looking for vague impressions and broad strokes rather than a specific answer.

src/localization/finnish.ts
```diff
@@ -25,61 +25,64 @@ export var finnishSurveyStrings = {
...
-  numericMax: "'{0}' täytyy olla vähemmän tai yhtä suuri kuin {1}",
+    "Luvun '{0}' tulee olla vähintään {1} ja korkeintaan {2}.",
+  numericMin: "Luvun '{0}' tulee olla vähintään {1}.",
+  numericMax: "Luvun '{0}' tulee olla korkeintaan {1}.",
   invalidEmail: "Syötä validi sähköpostiosoite.",
   invalidExpression: "Lausekkeen: {0} pitäisi palauttaa 'true'.",
   urlRequestError: "Pyyntö palautti virheen {0}. {1}",
   urlGetChoicesError:
-    "Pyyntö palautti tyhjät tiedot tai polku-ominaisuus on väärä",
+    "Pyyntö palautti tyhjän tiedoston tai 'path'-asetus on väärä",
```
vyznev, line 42: I'm not 100% sure in what contexts this message may appear, but based on the English text I'm assuming that there's a specific internal setting named `path` that may be wrong, and that it's not localized.  (At least there doesn't seem to be localized name for it.) Thus I assume it shouldn't be translated here, either.

vyznev (commented):
Added some notes on the translations, for the benefit of folks who don't speak Finnish. :)

andrewtelnov:
@vyznev Thank you!
Andrew

PR #2782, opened 2021-04-08
check this.arraysInfo 

swuecho:
can not set the row.value in matrix, without check `this.arrayInfo` first

ref: https://github.com/swuecho/survey-library/pull/1/files#diff-d842ef7dba53a6564a62405988f07c007898cd92b0445990b89ab3e6b2cce55dR206


tsv2013:
>can not set the row.value in matrix, without check this.arrayInfo first

It's very strange - we don't have any errors in our code. Probably array hasn't been properly created or initialized.

swuecho:
closed the request since not work on this part any more.

PR #2784, opened 2021-04-09
v-model="row.value"  is redundant in matrix `td` input

swuecho:
click event is handled by `td cellClick` .  It cause no problem in radio button, but when extend the input to `checkbox` will be trouble.


tsv2013:
Could you explain - how is this related to click event?

PR #2839, opened 2021-04-29
Serbian localization version 1

Vladimir-B2Bee:
serbian.ts file created for Serbian localization


tsv2013:
@Vladimir-B2Bee Thank you for the contribution

PR #2862, opened 2021-05-13
Update package.json

dmitrykurmanov:


dmitrykurmanov:
There was a bug in Testcafe release and now it should be ok

andrewtelnov (approved):

Issue #2682, opened 2021-02-27
Matrix allowRowsDragAndDrop


dmitry-kurmanov:
Matrix  `allowRowsDragAndDrop` property for drag and drop matrix rows

PR #3046, opened 2021-07-05
work for the #2682

dmitrykurmanov:


andrewtelnov (commented):
You can use a standard action bar item to create a drag&drop.
You will have to modify rendered table and add an action into left part if the property is set.
It will remove code duplication and you need to setup icon in the model only.
You will be able to add unit tests for this functionality in this case, by checking isActionsCell in renderedTable.

In function   "moveRowByIndex" value changing will happen two times, on removing and then adding row. You can get unbindValue from this.value, modify it and then set it as this.value = value;

PS: Lets do this functionality for matrix dynamic for now. It requires to write another code for matrix and matrix dropdown.

dmitrykurmanov:
could be merged

andrewtelnov (approved):

PR #3059, opened 2021-07-08
Feature/cv2 1453

OlgaLarina:


src/baseInterfaces.ts
```diff
@@ -0,0 +1,317 @@
+import { ILocalizableOwner, LocalizableString } from "./localizablestring";
```
tsv2013, line 1: Please rename this file to base-interfaces.ts

src/knockout/components/action-bar/action-bar.ts
```diff
@@ -26,25 +34,32 @@ export class ActionBarViewModel extends ActionBar {
...
     super(model);
 
-    model.items.forEach((item) => {
-      new ImplementorBase(item.stateItem);
+    ((<any>model).items || (<any>model).actions).forEach((item: any) => {
+      if (!!item.stateItem) {
+        new ImplementorBase(item.stateItem);
+      } else {
+        new ImplementorBase(item);
+      }
```
tsv2013, line 45: We need to wrap it with computed I think, otherwise we'll not update if actions will be modified

src/surveyError.ts
```diff
@@ -0,0 +1,32 @@
+import { ISurveyErrorOwner } from "./baseInterfaces";
```
tsv2013, line 1: survey-error.ts

src/utils/responsivitymanager.ts
```diff
@@ -1,4 +1,7 @@
+import { AdaptiveActionContainer } from "../actions/adaptive-container";
```
tsv2013, line 1: responsivity-manager.ts

tsv2013 (approved):

Issue #3072, opened 2021-07-09
Render question/panel collapse/expand button by using action bar


andrewtelnov:
We have a custom rendering for title collapse/expand action. It is time to replace it with our action bar.
Additionally, we should add tabindex and aria* attributes for header when there is collapse/expand button to allow expand/collapse panel and question content by keyboard

PR #3073, opened 2021-07-09
Render question/panel collapse/expand button by using action bar #3072

andrewtelnov:


src/action-bar.ts
```diff
@@ -238,6 +238,13 @@ export class AdaptiveElement extends Base {
   public get hasItems(): boolean {
     return (this.items || []).length > 0;
   }
+  public getItemById(id: string): IActionBarItem {
+    if (!this.items) return null;
+    for (var i = 0; i < this.items.length; i++) {
+      if (this.items[i].id === id) return this.items[i];
+    }
+    return null;
+  }
```
tsv2013, line 248: The src/action-bar.ts file contains deprecated code. We'll remove all the classes from this file. We need to use classes from the src/actions folder. I forgot to tell you.

andrewtelnov: Arh, OK. I have removed it. I wanted to use this function in Creator V2. Nevermind, I will add it later.

Thank you,
Andrew

tsv2013 (approved):

PR #3139, opened 2021-07-27
Accessibility updates 

andrewnoyes:
I didn't see an open issue for this, but if you'd like me to create one I'd be happy to. 

This PR addresses a few accessibility improvements:
* Adds `required` and `aria-required` to select components
* Updates errors, i.e. where role="alert", to include `aria-live="polite"` for screen readers
* Adds `role="img"` to svg icons


gologames (commented):
Notes about accessibility:
* The `role="img"` is necessary for proper announcing image in macOS Safari, VoiceOver (old versions) according to [Orange A11Y](https://a11y-guidelines.orange.com/en/articles/accessible-svg/)
* The `required` attribute for `select` tag is native. The `aria-required` is also recommended by [w3.org](https://www.w3.org/TR/WCAG20-TECHS/ARIA2.html) and [probably necessary](https://stackoverflow.com/a/39893468/4668668) for old browsers which don't support HTML5
* According to [CCC Accessibility Center](https://youtu.be/9zKT4GftZRc) `role="alert"` implicitly sets `aria-live="assertive"` which interrupts voice reader when appears. This behavior is annoying (according to [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions), so `aria-live="polite"` is set explicity

gologames (approved):

gologames:
Thank you for PR :sparkles:

Issue #3162, opened 2021-07-31
Return undefined for renderedMax for date instead of "2999-12-31"


andrewtelnov:
rendredMax property returns "2999-12-31" if max property is not set for text question with `inputType` equals to "date". It works fine for standard input, but breaks, for example bootstrap date picker that we support.
Returning `undefined` works for both widgets.

andrewtelnov:
We could not make this change, since we need 4 digits for year in the standard input and it is the only way to do it.

PR #3163, opened 2021-07-31
Return undefined for renderedMax for date instead of "2999-12-31" #3162

andrewtelnov:


tsv2013 (commented):
2999 date was used to limit date input for maximum 4 digits, otherwise it will allow enter 6 digits. We were asked for it before.

Issue #3187, opened 2021-08-10
Custom css styles from onUpdateQuestionCssClasses don't applied for T7487


ST-1580:
See [the ticket](https://surveyjs.answerdesk.io/ticket/details/T7487) in the support center

This problem appeared only in Survey Knockout because Knockout implementation calls `render` in constructor that cause calculating `cssClasses` value. Then when `onUpdateQuestionCssClasses` event set `cssClasses` is already calculated and its value doesn't update

ST-1580:
We've considered a number of approaches:
1) Issue occurs due to calculation of `cassClasses` on the [following line](https://github.com/surveyjs/survey-library/blob/111542df69c0cc5ac15cfb3fbfdc61c7e4b166c7/src/question.ts#L887):
```js
if (indent < 1 || !this.getSurvey() || !this.cssClasses) return "";
```
(in case of `startWithNewLine` equals false indent is greater then `0` and access to `cssClasses` lazy property cause it calculaton)

It is possible to change `cssClasses` to `cssClassesValue`, but some tests fails. Code is test set `indent` and check `paddingLeft` then. So anyway we should to calculate `cssClasses` at this place
2) Probably there is wrong render pipeline in SurveyKnockout (`render` method calls at the bottom of constructor). But also a lot of tests expect that some of "ko" values is calculated after instantiation of Survey Knockout
3) As about invalidate `cssClassesValue` hash after setting `onUpdateQuestionCssClasses` event there is no built-in API to override `add` method of `EventBase` and we didn't found examples of this approach in code
4) As a result we added invalidating of `cssClassesValue` value via `this.updateElementCss(false);` call in `render` method of SurveyKnockout

PR #3188, opened 2021-08-10
Custom css styles from onUpdateQuestionCssClasses don't applied for #3187

ST-1580:
See more information in [the issue](https://github.com/surveyjs/survey-library/issues/3187)


andrewtelnov (approved):
I am not sure, why do you update other files rather than kosurvey.ts?

gologames:
> I am not sure, why do you update other files rather than kosurvey.ts?

Other changes are just replacing `var` to `const`/`let`, removing unnecessary imports and so on. I understand that this is not directly related to the issue, but I have no better idea how to refactor things like these

PR #3190, opened 2021-08-12
doPropertyValueChangedCallback mismatch

Johnnyrook777:
The `doPropertyValueChangedCallback` function is checked, but `onPropertyValueChangedCallback` was called.

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -465,7 +465,7 @@ export class Base {
     if (!target) target = this;
     var notifier: any = this.getSurvey();
     if (!notifier) notifier = this;
-    if (!!notifier.doPropertyValueChangedCallback) {
+    if (notifier !== this && !!notifier.doPropertyValueChangedCallback) {
       notifier.doPropertyValueChangedCallback(
         name,
         oldValue,
--- a/src/base.ts
+++ b/src/base.ts
@@ -466,7 +466,7 @@ export class Base {
     var notifier: any = this.getSurvey();
     if (!notifier) notifier = this;
     if (!!notifier.doPropertyValueChangedCallback) {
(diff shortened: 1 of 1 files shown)


andrewtelnov:
@Johnnyrook777 You are using an old version. We have a different code now. There is indeed a typo that I have just fixed, but it is another issue. Change the check from [!!notifier.doPropertyValueChangedCallback)](https://github.com/surveyjs/survey-library/blob/master/src/base.ts#L468) to `!!notifier.onPropertyValueChangedCallback)`.
All others is correct. We call `onPropertyValueChangedCallback` and not `doPropertyValueChangedCallback`.

Thank you,
Andrew

andrewtelnov:
You can check the base.ts file history.

Thank you,
Andrew

Johnnyrook777:
Thanks @andrewtelnov . 
Im not really familiar with this code base, but at a glance it appears that the `onPropertyValueChangedCallback` method will be called multiple times. 
Do you want me to continue on with this PR? The code changes you have made appear to fix the issue I was having, and Im a bit reluctant to add this code now as I dont understand the code base well enough. 

Regards

andrewtelnov:
@Johnnyrook777 There are several unit tests that check this functionality, including number of calls.
https://github.com/surveyjs/survey-library/blob/master/tests/basetests.ts#L393
https://github.com/surveyjs/survey-library/blob/master/tests/calculatedvaluestests.ts#L309
It calls the onPropertyValueChangedCallback for survey and for the object itself, if onPropertyValueChangedCallback callback is set.

There is nothing to fix.

Thank you,
Andrew

Johnnyrook777:
Excellent. Thanks Andrew

PR #3204, opened 2021-08-20
https://github.com/surveyjs/survey-creator/issues/1613

OlgaLarina:


src/actions/adaptive-container.ts
```diff
@@ -28,6 +30,44 @@ export class AdaptiveActionContainer<T extends Action = Action, V extends IActio
       });
   }
 
+  private showFirstN(visibleItemsCount: number) {
+    let leftItemsToShow = visibleItemsCount;
```
ST-1580, line 34: Is that by design that `visibleItemsCount` is unused? It is possible to use `visibleItemsCount`  instead `leftItemsToShow`

src/actions/adaptive-container.ts
```diff
@@ -28,6 +30,44 @@ export class AdaptiveActionContainer<T extends Action = Action, V extends IActio
...
   }
 
+  private showFirstN(visibleItemsCount: number) {
+    let leftItemsToShow = visibleItemsCount;
+    const invisibleItems: Action[] = [];
+    this.actions.filter(action => action.visible).forEach((item) => {
+      if (item === this.dotsItem) {
+        return;
+      }
+      if (leftItemsToShow <= 0) {
```
ST-1580, line 40: Elements that index greater then `visibleItemsCount` carry to `invisibleItems` and its `mode` set to `"popup"`. But there is no actions with elements that index less then `visibleItemsCount`. Is it OK? Anyway method name is unobvious

src/actions/adaptive-container.ts
```diff
@@ -28,6 +30,44 @@ export class AdaptiveActionContainer<T extends Action = Action, V extends IActio
...
+        item.mode = "popup";
+        invisibleItems.push(item);
+      }
+      leftItemsToShow--;
+    });
+    this.invisibleItemsListModel.items = invisibleItems;
+  }
+
+  private getVisibleItemsCount(size: number): number {
```
ST-1580, line 49: Probably exact name for `size` parameter is `availableSize`?

src/actions/adaptive-container.ts
```diff
@@ -28,6 +30,44 @@ export class AdaptiveActionContainer<T extends Action = Action, V extends IActio
...
+    const itemsSizes: number[] = this.visibleActions.map((item) => item.minDimension);
+    let currSize: number = 0;
+    for (var i = 0; i < itemsSizes.length; i++) {
+      currSize += itemsSizes[i];
+      if (currSize > size) return i;
+    }
+    return i;
+  }
+
+  private updateItemMode(availableSize: number, itemsSize: number) {
```
ST-1580, line 59: Perhaps is makes sense to rename `itemsSize` property name to `maxSize`?

ST-1580 (approved):

PR #3251, opened 2021-09-02
Only apply aria-labelledby if an aria-role is set

andrewnoyes:
From [2.10 Practical Support: aria-label, aria-labelledby and aria-describedby](https://www.w3.org/TR/using-aria/#label-support): 

> Don't use aria-label or aria-labelledby on a span or div **_unless_** its given a role. When aria-label or aria-labelledby are on interactive roles (such as a link or button) or an img role, they override the contents of the div or span. Other roles besides Landmarks (discussed above) are ignored.

I updated each question template to only set this attribute if a role was also provided

Before and after test results using the [ARC toolkit](https://chrome.google.com/webstore/detail/arc-toolkit/chdkkkccnlfncngelccgbgfmjebmkmce?hl=en):

Screen Shot 2021-09-02 at 12 17 30 PM

Screen Shot 2021-09-02 at 12 15 52 PM


dmitrykurmanov:
Hello,

thanks for the pull request I agree with it. But after the investigation I've realized that we don't use the `question.ariaRole` for all our questions. In some cases we use inline html attribute like `<div role="radiogroup">`. 

So I am going to fix it and do some other refactoring and then I will approve your PR. The issue #3258.

Thank you,
Dmitry
SurveyJS Team

dmitrykurmanov:
Thanks for the PR!

dmitrykurmanov (approved):

dmitrykurmanov:
@andrewnoyes  hello, Thanks again for the PR we will apply it a bit later

@andrewtelnov reverted the pull request because of the functional tests I've missed. We have a lot of tests depends on this functionality. We are working on accessibility for now and I am going to apply your changes, fix functional tests, and do some refactoring here in the issue https://github.com/surveyjs/survey-library/issues/3258.

Issue #3343, opened 2021-09-17
Async expression validators don't focus element when false at validation


FHourmy:
### Are you requesting a feature, reporting a bug or asking a question?
Bug (maybe feature ?)

### What is the current behavior?
In an async expression validator if the expression is false, survey won't scroll to the element on validation

### What is the expected behavior?
On validation, the page should scroll up to the element with the invalid expression validator

### How would you reproduce the current behavior (if this is a bug)?
create a survey with an element with an expression validator calling an async function with this.returnResult(false);,
scroll down the page, press the complete button.

#### Provide the test code and the tested page URL (if applicable)
https://codesandbox.io/s/sparkling-flower-tdkbq?file=/index.js

### Specify your

* browser: chrome / FF
* browser version:
* surveyjs platform (angular or react or jquery or knockout or vue): jquery
* surveyjs version: 1.8.67

andrewtelnov:
@FHourmy I guess, we do not focus question if errors come from async. I will check and come back.

Thank you,
Andrew

andrewtelnov:
@FHourmy It is fixed by the commit above.

Thank you,
Andrew

PR #3352, opened 2021-09-20
Fix: Async expression validators don't focus element when false at va…

andrewtelnov:
…lidation #3343

### DIFF
--- a/tests/surveytests.ts
+++ b/tests/surveytests.ts
@@ -12648,6 +12648,55 @@ QUnit.test(
   }
 );
 QUnit.test(
+  "Focus question on Survey onServerValidateQuestions event, Bug#3343",
+  function (assert) {
+    var focusedQuestionId = "";
+    var oldFunc = SurveyElement.FocusElement;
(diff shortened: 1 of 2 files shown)


tests/surveytests.ts
```diff
@@ -12647,6 +12647,55 @@ QUnit.test(
...
+  "Focus question on Survey onServerValidateQuestions event, Bug#3343",
+  function (assert) {
+    var focusedQuestionId = "";
+    var oldFunc = SurveyElement.FocusElement;
+    SurveyElement.FocusElement = function (elId: string): boolean {
+      focusedQuestionId = elId;
+      return true;
+    };
+    var returnResult: (res: any) => void;
+    function asyncFunc(params: any): any {
```
dmitrykurmanov, line 12660: The function `asyncFunc` looks sync. Should we use `setTimeout` here?

andrewtelnov: I use callback function to emulate async in sync unit test

dmitrykurmanov (approved):

Issue #3363, opened 2021-09-22
Introduce createItemValue into base select class


andrewtelnov:
We need "createItemValue" function that will create a correct item value to avoid code duplication in Library and Creator V2.
This function only creates a correct ItemValue class and do not add it into choices.

PR #3364, opened 2021-09-22
Introduce createItemValue into base select class #3363

andrewtelnov:


src/question_baseselect.ts
```diff
@@ -92,6 +95,12 @@ export class QuestionSelectBase extends Question {
     }
     this.removeFromDependedQuestion(this.getQuestionWithChoices());
   }
+  protected getItemValueType() {
+    return "itemvalue";
+  }
```
tsv2013, line 100: We have the item value type. Probably we could create it via serializer in base item value class...

tsv2013 (approved):

PR #3392, opened 2021-09-28
Feature/cv2 1768

dk981234:


src/question_file.ts
```diff
@@ -434,17 +436,26 @@ export class QuestionFileModel extends Question {
...
     }
+    this.isDragging = true;
     event.dataTransfer.dropEffect = "copy";
     event.preventDefault();
   }
   onDrop = (event: any) => {
-    event.preventDefault();
-    let src = event.dataTransfer;
-    this.onChange(src);
+    if (!this.isReadOnly && ! this.isInputReadOnly) {
```
tsv2013, line 448: Can we add a unit test for read only mode?

tsv2013 (approved):

PR #3415, opened 2021-10-04
Bug/1925 string editor validation

novikov82:


src/expressionItems.ts
```diff
@@ -39,6 +39,9 @@ export class ExpressionItem extends Base implements ILocalizableOwner {
   public getRenderer(name: string): string {
     return !!this.locOwner ? this.locOwner.getRenderer(name) : null;
   }
+  public getRendererContext(locStr: LocalizableString): string {
```
tsv2013, line 42: Probably we don't need them

Issue #3419, opened 2021-10-04
Accessibility aria-label in legend tag


ST-1580:
For more detailed information see [the ticket in support center](https://surveyjs.answerdesk.io/ticket/details/T7877)

PR #3420, opened 2021-10-04
Work for #3419 - added role to legend tag

ST-1580:


dmitrykurmanov (commented):
It would be better to set up the `role` in the top-level `question` element. See: https://github.com/surveyjs/survey-library/blob/a1d521a70ae2124ce69102c901de19e74c4c051f/src/react/reactquestion.tsx#L127
But it works only for [radiogroup](https://github.com/surveyjs/survey-library/blob/a1d521a70ae2124ce69102c901de19e74c4c051f/src/question_radiogroup.ts#L45) for now. 

So we only need to add something like this:
```ts
  public get ariaRole(): string {
    return "radiogroup";
  }
```
to the checkbox and it will be the fix. But we need to investigate what exactly role it should be.
And also setup correct roles for the items too. [Example](https://github.com/surveyjs/survey-library/blob/a1d521a70ae2124ce69102c901de19e74c4c051f/src/react/reactquestion_radiogroup.tsx#L160).

I will investigate the issue a bit later and we could discuss it.

dmitrykurmanov:
discussed with @ST-1580 and created the issue https://github.com/surveyjs/survey-library/issues/3421 to implement in the future

dmitrykurmanov (approved):

PR #3457, opened 2021-10-08
Bug/cv2 1973 return animation choices ranking

dmitry-kurmanov:
https://github.com/surveyjs/survey-creator/issues/1973


dmitrykurmanov:
> We need to find a way to test all this functionality

yes, maybe we could check properly animation class existance after some drag actions. With testcafe it should be possible. I'll create an issue.

PR #3568, opened 2021-11-02
fix active page in design mode

novikov82:
surveyjs/survey-creator#2147


novikov82:
@andrewtelnov @tsv2013 
I could fix the bug, but really can not explain how I did it.
I don't know why isStatedSet is set in Creator, so I needed to cover it with isDesignMode flag.
Maybe there is more clear and understandable solution for this?

tsv2013 (approved):

PR #3574, opened 2021-11-03
matrix dynamic total calc

### LINKED ISSUE #3565: Dynamic Matrix Totals do not work in jQuery
Reporting a bug.

Adding a totals row to a dynamic matrix produces a working total row in React. However, the same configuration in jQuery does not show the totals row when data is changed.

The totals row should display in jQuery.

Create a survey for use with jQuery SurveyJS, with a dynamic matrix question with an input type of number. Add a totals row with total type "sum". Enter a value in the field. The totals row disappears.

#### Provide the test code and the tested page URL (if applicable)

Tested page URL (React, working - change a value to check): https://plnkr.co/edit/ac4QL6tSPWrByH8Q
Tested page URL (jQuery, not working - change a value to check): https://plnkr.co/edit/jldAIrxUvIjguLYv

### Specify your

* browser: Chrome
* browser version: 95.0.4638.54
* surveyjs platform (angular or react or jquery or knockout or vue): jQuery
* surveyjs version: 1.8.75 (NOTE: this worked in 1.8.70 and stopped working in 1.8.71)

novikov82:
fixed #3565

### DIFF
--- a/src/knockout/templates/question-expression.html
+++ b/src/knockout/templates/question-expression.html
@@ -1,3 +1,3 @@
 <script type="text/html" id="survey-question-expression">
-  <div data-bind="text:question.displayValue, css: question.koCss().root"></div>
+  <div data-bind="text:question.koValue, css: question.koCss().root"></div>
 </script>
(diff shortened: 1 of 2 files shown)


andrewtelnov (commented):
expression displayValue is a correct property here. value will shows without formating, for example currency symbols. So the fix is incorrect. We need to let know knockout that displayValue has been changed. We likely need to introduce another property that will be updated on setting value or changing formating.
It could be calculated property. Let me think.

Andrew

PR #3587, opened 2021-11-09
v2: DnD: while dragging selected element the "selection"(orange border) should go to the shortcut and moved to element after drag 

dmitrykurmanov:
see https://github.com/surveyjs/survey-creator/issues/2172

### DIFF
--- a/src/dragdrop/core.ts
+++ b/src/dragdrop/core.ts
@@ -130,26 +130,12 @@ export abstract class DragDropCore<T> extends Base {
   protected createDraggedElementShortcut(text: string, draggedElementNode?: HTMLElement, event?: PointerEvent): HTMLElement {
     const draggedElementShortcut = document.createElement("div");
     draggedElementShortcut.innerText = text;
-    draggedElementShortcut.style.cssText = this.getDraggedElementShortcutStyles();
+    draggedElementShortcut.className = this.getDraggedElementClass();
     return draggedElementShortcut;
   }
 
-  protected getDraggedElementShortcutStyles() {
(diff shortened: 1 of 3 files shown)


src/dragdrop/core.ts
```diff
@@ -130,11 +130,28 @@ export abstract class DragDropCore<T> extends Base {
   protected createDraggedElementShortcut(text: string, draggedElementNode?: HTMLElement, event?: PointerEvent): HTMLElement {
     const draggedElementShortcut = document.createElement("div");
     draggedElementShortcut.innerText = text;
-    draggedElementShortcut.style.cssText =
-      "height: 24px; min-width: 100px; border-radius: 36px; background-color: white; padding: 16px; cursor: grabbing; position: absolute; z-index: 1000; box-shadow: 0px 8px 16px rgba(0, 0, 0, 0.1); font-family: 'Open Sans'; font-size: 16px; padding-left: 20px; line-height: 24px;";
+    draggedElementShortcut.style.cssText = this.getDraggedElementShortcutStyles();
     return draggedElementShortcut;
   }
 
+  protected getDraggedElementShortcutStyles() {
```
andrewtelnov, line 137: Is it possible to define a CSS rule here?
It should support our default color and we should give the ability to override attributes.

dmitrykurmanov: it make sense I'll fix it

dmitrykurmanov: fixed

andrewtelnov (approved):

PR #3626, opened 2021-11-18
Font-family not latin characters

ST-1580:
For detailed information see https://github.com/surveyjs/survey-creator/issues/2090

### DIFF
--- a/src/defaultV2-theme/variables.scss
+++ b/src/defaultV2-theme/variables.scss
@@ -34,7 +34,7 @@ $green: var(--green, #19b394);
 $green-light: var(--green-light, rgba(25, 179, 148, 0.1));
 $blue-light: var(--blue-light, rgba(67, 127, 217, 0.1));
 
-$font-family: "Open Sans", "Roboto", sans-serif;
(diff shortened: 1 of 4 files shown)


src/defaultV2-theme/variables.scss
```diff
@@ -34,7 +34,7 @@ $green: var(--green, #19b394);
 $green-light: var(--green-light, rgba(25, 179, 148, 0.1));
 $blue-light: var(--blue-light, rgba(67, 127, 217, 0.1));
 
-$font-family: "Open Sans";
+$font-family: "Open Sans", "Roboto", sans-serif;
```
tsv2013, line 37: "Roboto" is non-standard. We need to use: "Helvetica Neue", Helvetica, Arial, sans-serif
After open sans

tsv2013 (commented):
We need to remove non-EU open sans fonts

tsv2013 (changes requested):
We need to remove non-latin fonts

tsv2013 (approved):

Issue #3635, opened 2021-11-19
Ranking shortcut: add css class


dmitrykurmanov:
Need for the customization. See https://surveyjs.answerdesk.io/ticket/details/T8211

PR #3636, opened 2021-11-19
Ranking shortcut: add css class

dmitrykurmanov:
see https://github.com/surveyjs/survey-library/issues/3635


tsv2013 (changes requested):
It looks like we can add a unit test for this case

dmitrykurmanov:
> It looks like we can add a unit test for this case

fixed

tsv2013 (approved):

PR #3641, opened 2021-11-19
Update react peer dep to 17

LeifAndersen:
Would this be possible given that survey-react-ui uses react 17?

### DIFF
--- a/build-scripts/survey-react/package.json
+++ b/build-scripts/survey-react/package.json
@@ -28,7 +28,7 @@
   },
   "typings": "survey.react.d.ts",
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


tsv2013 (changes requested):
I think according the https://docs.npmjs.com/cli/v8/configuring-npm/package-json it should be something like
">16.5.0 <18.0.0" or "^16.5.0 || ^17.0.0"

tsv2013:
We've fixed it via the https://github.com/surveyjs/survey-library/pull/3713/commits/0a6fab15f09fbfefc222778c80e849d3ecb0724a commit

PR #3716, opened 2021-12-08
Fix incorrectly spelt property value

breezedave:


src/question.ts
```diff
@@ -1439,7 +1439,7 @@ export class Question extends SurveyElement
   protected updateIsAnswered(): void {
     const oldVal = this.isAnswered;
     this.setPropertyValue("isAnswered", this.getIsAnswered());
-    if(oldVal !== this.isAnswered) {
+    if (oldVal !== this.isAnswered) {
```
tsv2013, line 1442: Please remove space between "if" and "("

tsv2013 (changes requested):

src/question.ts
breezedave: Removed. 

The rest of the file has a space after its "if" statements. Which is the preferred format for the library?

andrewtelnov (approved):

tsv2013 (approved):

tsv2013:
@breezedave  Thank you for the cotributing!

Issue #3772, opened 2021-12-16
Accessibility violations for checkboxes


dszewczyk1:
### Are you requesting a feature, reporting a bug or asking a question?
Question/bug

### What is the current behavior?
Violation reports based on IBM Equal Access Accessibility Checker (Chrome extension) https://chrome.google.com/webstore/detail/ibm-equal-access-accessib/lkcagbfjnkomcinoddgooolagloogehp)
> `<fieldset>` element `<legend>` is empty
> An element with ARIA role 'checkbox' does not have the required ARIA attribute(s): 'aria-checked'
> The ARIA role 'checkbox' is not valid for the element `<legend>`

And that's because of the `<legend>` added like this:
`<legend role="checkbox" aria-label="What car are you driving?"></legend>`

I can see the reasoning that you've added aria-label instead of providing a text (in other way, you'd need to visually hide it) 
But why `<legend>` even has role="checkbox"?

It's also complaining about
> Accessible name does not match or contain the visible label text

I guess it doesn't like the aria-label here, which doesn't match "Ford"
`<input value="Ford" aria-invalid="false" aria-label="What car are you driving?" aria-required="true" id="sq_100i_0" name="car" type="checkbox" class="sv-visuallyhidden sv-item__control">`

### What is the expected behavior?
No violations reported

### How would you reproduce the current behavior (if this is a bug)?
Run IBM Equal Access Accessibility Checker scan on a page with checkboxes

#### Provide the test code and the tested page URL (if applicable)

Tested page URL: https://surveyjs.io/Examples/Library?id=questiontype-checkbox&platform=Reactjs&theme=modern

### Specify your

* browser: Chrome
* browser version: 96.0.4664.93
* surveyjs platform (angular or react or jquery or knockout or vue): all, I think (but React in my case)
* surveyjs version: 1.9.2

dmitrykurmanov:
We will investigate the problem and write back asap

dmitrykurmanov:
I believe that we fixed the most problems here. At least nvda reader sounds ok now. Fix will be available in the next minor release.

As for `fieldset` and `legend` tags we don't use them for the accessibility because of we have a very custom and complex markup for the questions. We set `role="presentation"` for them (IBM checker doesn't respect the `role="presentation"` for some reasons). That's why we use custom `aria-labels`.

dszewczyk1:
[...]

PR #3868, opened 2022-01-13
Accessibility violations for checkboxes

dmitrykurmanov:
see https://github.com/surveyjs/survey-library/issues/3772

### DIFF
--- a/src/knockout/templates/question-checkbox.html
+++ b/src/knockout/templates/question-checkbox.html
@@ -1,5 +1,6 @@
 <script type="text/html" id="survey-question-checkbox">
   <fieldset data-bind="css: question.koCss().root" role="presentation">
+      <legend role="presentation" class="sv-hidden"></legend>
       <!-- ko ifnot: question.hasColumns  -->
         <!-- ko foreach: { data: question.visibleChoices, as: 'item', afterRender: question.koAfterRender }  -->
           <!-- ko component: { name: question.getItemValueWrapperComponentName(item), params: { componentData:  question.getItemValueWrapperComponentData(item), templateData: { name: 'survey-checkbox-item', data: item } } } -->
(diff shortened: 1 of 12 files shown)


dmitrykurmanov:
@dszewczyk1 I checked it yesterday and didn't see this message. I'll check it again.

dmitrykurmanov:
> is the `legend` supposed to be empty?

I can't see that errors now:
image

we have only one: "The ARIA attribute 'aria-required, aria-invalid' is not valid for the element "div" with ARIA role 'group'"
but it seems that checker has outdated check. Because in the referenced [doc ](https://www.w3.org/TR/html-aria/#docconformance) we could see the following:

><html>
<body>


div | No corresponding role | Any roleGlobal aria-* attributes and any aria-* attributes applicable to the allowed roles.
-- | -- | --



</body>
</html>

dszewczyk1:
@dmitrykurmanov according to this https://www.w3.org/TR/wai-aria-1.2/#aria-invalid `aria-invalid` and `aria-required` are allowed for specific roles only and probably they're using it for this checker.
`This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.`
Similar violations for aria-invalid and aria-required are shown in the other types, for example: https://surveyjs.io/Examples/Library?id=questiontype-text&platform=Reactjs&theme=modern

dmitrykurmanov:
> @dmitrykurmanov according to this https://www.w3.org/TR/wai-aria-1.2/#aria-invalid `aria-invalid` and `aria-required` are allowed for specific roles only and probably they're using it for this checker. `This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.` Similar violations for aria-invalid and aria-required are shown in the other types, for example: https://surveyjs.io/Examples/Library?id=questiontype-text&platform=Reactjs&theme=modern

Probably we should choose another role for our "checkbox group". Or find some "checkboxes group" example... Here the separate issue: https://github.com/surveyjs/survey-library/issues/3874

tsv2013 (approved):

PR #3892, opened 2022-01-19
add supported validators for the image picker

OlgaLarina:


dmitrykurmanov (changes requested):
probably we should add unit test : https://github.com/surveyjs/survey-library/blob/57eac3ac4328137f4cad6bbbde648d44c456515f/tests/surveyquestiontests.ts#L3847

OlgaLarina:
@dmitrykurmanov 
I have added a unit test.

dmitrykurmanov (approved):

PR #3957, opened 2022-02-01
Resolved #3850 - Tooltip not working as expected with markdown and matrix - removed tooltip

### LINKED ISSUE #3850: Tooltip not working as expected with markdown and matrix

bug/unexpected feature

The artefact can be seen by e.g. mouse over the dog. the html tags appears as text in the tooltip/title when hovering.

https://surveyjs.io/Examples/Library?id=survey-markdown-radiogroup&platform=jQuery&theme=bootstrap

it would be nice to specify separately what should be presented as tooltip and what should be presented as (html) option. 

e.g. for an option specify 

```
{
    name : 'id1',
    text : 'as it is today, used where other options does not make sense',
    markDown : 'feed into markdown where html would be nice', 
    html : '<b>preformatted</b> html code as alternative to markDown',
    otherInfo : 'markdown should receive the entire object, so even this property could',
    otherinfo2: 'be used for rendering',
}

```
(see link above)

### Specify your

* browser: chrom, opera
* browser version: 82
* surveyjs platform (angular or react or jquery or knockout or vue): jquery bootstrap 
* surveyjs version: 9.

tsv2013:


src/knockout/templates/question-imagepicker.html
```diff
@@ -22,7 +22,7 @@
                   <!-- /ko -->
                 </div>
                 <!-- ko if: question.showLabel -->
-                <span data-bind="css: question.koCss().itemText, attr: { title: item.locText.koRenderedHtml }">
+                <span data-bind="css: question.koCss().itemText">
```
andrewtelnov, line 25: I an not sure that we have to remove title from image picker. Probaby it makes sense to have it.

Andrew

tsv2013: We needn't a title here as well because text is visible - https://github.com/surveyjs/survey-library/blob/master/src/knockout/templates/question-imagepicker.html#L25-L28

andrewtelnov (approved):

PR #4137, opened 2022-03-22
Redesign list search in popups

OlgaLarina:


novikov82 (commented):
What about visual regression test?

tsv2013 (approved):

PR #4206, opened 2022-04-06
Added theme markers for https://github.com/surveyjs/survey-creator/issues/2844

ST-1580:


tsv2013 (changes requested):
Why wouldn't us to use theme name CSS variable? It will be set in different value in each theme.
They can be accessed via the following code:
```JS
var bodyStyles = window.getComputedStyle(document.body);
var fooBar = bodyStyles.getPropertyValue('--foo-bar'); //get
document.body.style.setProperty('--foo-bar', newValue);//set
```

src/defaultV2-theme/defaultV2.scss
```diff
@@ -50,6 +50,10 @@
 @import "../signaturepad.scss";
 @import "./defaultV2.m600.scss";
 
+body {
+  --sd-theme-mark: true;
```
tsv2013, line 54: May be it should be --sv_theme_mark ?

dk981234 (approved):

tsv2013 (approved):

Issue #4275, opened 2022-04-27
element title collapse/expand action availability


dk981234:
- [x] add focus state for element's title (toggleState, focusFirst input actions)
- [x] make whole header with description clickable when toggling state (now works only focusFirstInput)

PR #4276, opened 2022-04-27
Work for #4275: expand/collapse action usability 

dk981234:
Make whole header clickable for toggle state action. Add focus styles for title of collapsed question


tsv2013 (approved):

andrewtelnov (commented):
Everything is fine to me.

PR #4343, opened 2022-05-13
Work for #4257 - questionCount 1 or greater for progressBar calculation

joeyskuo:
fix for progressBar value defaulting to 100% for unanswered 1-question surveys. modified getProgress method to calculate progress percentage when question count is equal to 1


tsv2013 (approved):

tsv2013:
@joeyskuo  Thank you for the contribution!

PR #4366, opened 2022-05-23
Add into survey base object isDescendantOf function fixed #4365

### LINKED ISSUE #4365: Add into survey base object isDescendantOf function
We have getType() function that returns the object type.
Sometimes you need to know if the object instance is inherited from another object.
For example, is the question  "dropdown" or inherited from "dropdown". In this case `obj.getType() === "dropdown"` will not work and it should be used `obj.isDescendantOf("dropdown");` in this case.

andrewtelnov:


src/base.ts
```diff
@@ -302,6 +302,15 @@ export class Base {
   public getType(): string {
     return "base";
   }
+  /**
+   * Use it to find out if the current instance has getType() equals to 'typeName' or inherited from this typeName
+   * @param typeName the string in lowercase, a SurveyJS type name
+   * @returns true if the current instance has getType() equals to 'typeName' or inherited from this typeName
+   * @see getType
```
RomanTsukanov, line 309:
```suggestion
   * Use this method to find out if the current instance is of the given `typeName` or inherited from it.
   * @param `typeName` One of the values listed in the [getType()](https://surveyjs.io/Documentation/Library?id=surveymodel#getType) description.
   * @returns true if the current instance is of the given `typeName` or inherited from it
   * @see getType
```

dirk-pieterse (approved):

PR #4383, opened 2022-05-27
resolve #4382 Dropdown render as select: popup has narrow width

### LINKED ISSUE #4382: Dropdown render as select: popup has narrow width
image

OlgaLarina:


tsv2013 (changes requested):
We need to get rid of this code duplication
```TS
    if (!!event && !!event.target) {
      const target = findParentByClass(event.target, this.question.cssClasses.control);
      if (!!target) {
        PopupUtils.updatePopupWidthBeforeShow(this.question.popupModel, target, event);
      }
    }
```

tsv2013 (approved):

PR #4496, opened 2022-06-23
Dropdown accessibility

OlgaLarina:


dmitrykurmanov (commented):
probably we also need a markup tests for the changes

tsv2013 (approved):

PR #4597, opened 2022-07-18
Add the Conditional Logic and Dynamic Texts help topic

RomanTsukanov:


docs/design-survey-conditional-logic.md
```diff
@@ -0,0 +1,692 @@
...
+    "setValue": "adult"
+  }]
+}
+```
+
+---
+
+### `copyvalue`
+
+Copies a value from one question to another. The `fromName` property specifies the source question; the `setToName` property specifies the target question.
```
dmitry-eliseev-devexpress, line 660:
```suggestion
Copies a value from between questions. The `fromName` property specifies the source question; the `setToName` property specifies the target question.
```

docs/design-survey-conditional-logic.md
```diff
@@ -0,0 +1,692 @@
...
+    "setToName": "shippingAddress"
+  }]
+}
+```
+
+---
+
+### `runexpression`
+
+If the `expression` is `true`, runs another expression specified by the `runExpression` property. You can also save the result of `runExpression` as a question value. For this, assign the question's name to the `setToName` property.
```
dmitry-eliseev-devexpress, line 690: > If the `expression` is `true`, **runs** another expression specified by the `runExpression` property.

Who runs?

docs/design-survey-conditional-logic.md
```diff
@@ -0,0 +1,692 @@
...
+- **Boolean expression**    
+  An expression that evaluates to `true` or `false`. Boolean expressions are widely used to implement conditional logic. Refer to the following help topic for more information: [Conditional Visibility](#conditional-visibility).
+
+Expressions can include question names, variables, and calculated values (described in the [Dynamic Texts](#dynamic-texts) section). Plus expressions can use [built-in](#built-in-functions) and [custom functions](#custom-functions).
+
+### Built-In Functions
+
+Functions allow you to perform additional calculations within an expression. One expression can contain multiple function calls.
+
+Functions can accept arguments. For example, the following expression shows the built-in [`age`](#age) and [`iif`](#iif) functions. `age` accepts the value of a `birthdate` question. `iif` accepts three arguments: a condition, a value to return when the condition is truthy, and a value to return when the condition is falsy.
```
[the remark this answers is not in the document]
RomanTsukanov: [Truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) and [falsy](https://developer.mozilla.org/en-US/docs/Glossary/falsy) are legit terms in JavaScript world.

docs/design-survey-conditional-logic.md
RomanTsukanov: The trigger obviously. This is the same pattern used in the previous descriptions of triggers, just with an if-part.

docs/design-survey-conditional-logic.md
```diff
@@ -0,0 +1,691 @@
...
+
+### Calculated Values
+
+Calculated values allow you to register an [expression](#expressions) under a required name. If the expression includes [questions](#question-values), [variables](#variables), or [functions](#built-in-functions), it is recalculated each time their values are changed.
+
+To configure a calculated value, define the [`calculatedValues`](https://surveyjs.io/Documentation/Library?id=surveymodel#calculatedValues) array in the survey JSON definition. Each object in this array should contain the following fields:
+
+- `name` - A name that identifies the calculated value.
+- `expression` - An expression that returns the calculated value.
+- `includeIntoResult` - A Boolean property that specifies whether to include the calculated value into survey results.
```
annnke, line 115: `includeIntoResult` - A Boolean property that specifies whether to include the calculated value in survey results.

docs/design-survey-conditional-logic.md
```diff
@@ -0,0 +1,691 @@
...
+  // Preserve the context variable to access the `returnResult` callback later
+  var self = this;
+  setTimeout(function() {
+    // Return the function result via the callback
+    self.returnResult(yourValue);
+  }, 100);
+}
+```
+
+After you implement an asynchronous function, register it in the `FunctionFactory`. The third parameter specifies if this function is asynchronous:
```
annnke, line 486: After you implement an asynchronous function, register it in `FunctionFactory`. The third parameter specifies if this function is asynchronous:

docs/design-survey-conditional-logic.md
```diff
@@ -0,0 +1,691 @@
...
+    ]
+  },
+  // ...
+  ]
+}
+```
+
+[View example](https://surveyjs.io/Examples/Library?id=condition-choicesVisibleIf)
+
+This technique has one drawback: if a question contains many items, you have to copy the same expression into every item that should have dynamic visibility. If that is your case, use a technique described in the next topic.
```
annnke, line 563: This technique has one drawback: if a question contains many items, you have to copy the same expression into every item that should have dynamic visibility. If that is your case, use the technique described in the next topic.

docs/design-survey-conditional-logic.md
```diff
@@ -0,0 +1,691 @@
...
+    "setToName": "shippingAddress"
+  }]
+}
+```
+
+---
+
+### `runexpression`
+
+If the `expression` is `true`, runs another expression specified by the `runExpression` property. You can also save the result of `runExpression` as a question value. For this, assign the question's name to the `setToName` property.
```
annnke, line 689: If the `expression` is `true`, the trigger runs another expression specified by the `runExpression` property. You can also save the result of `runExpression` as a question value. For this, assign the question's name to the `setToName` property.

annnke (approved):

Issue #4696, opened 2022-08-08
Panels with high nesting level look bad


RomanTsukanov:
![image](https://user-images.githubusercontent.com/18551316/183411678-86dfe329-e976-4f74-997b-f83a0f245d62.png)
![image](https://user-images.githubusercontent.com/18551316/183411698-94be400e-bb5f-4915-9890-3054e8748eb1.png)

You can reproduce the issue in this plunk: https://plnkr.co/edit/k3VFs4C3NBcUdBHW
Select "Married", then select "Yes" and "Other types of employment" for both you and your partner.

dk981234:
- [ ] Fix for desktop
- [ ] Fix for small screens

PR #4763, opened 2022-08-25
Work for #4696:  fix in case of rating

dk981234:


dk981234:
Seems to be working with "default" and "modern" theme too

tsv2013 (approved):

PR #4855, opened 2022-09-15
Add locTitleName and locTooltipName properties into IAction interface…

### LINKED ISSUE #4854: Add locTitleName and locTooltipName properties into IAction interface
You can use these properties to get strings from localization. For example:
```
survey.addNavigationItem({
    id: "action1",
    locTitleName: "selectAllItemText",
    locToolTipName: "previewText"
  });
```
"selectAllItemText" and "previewText" the name of the translated strings: "Select All" and "Preview" in English, "Alles auswählen" and "Vorschau" in Deutsch and so on.

andrewtelnov:
… fixed #4854

### DIFF
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -242,6 +242,17 @@ export class Action extends Base implements IAction, ILocalizableOwner {
   public locStrsChanged(): void {
     super.locStrsChanged();
     this.locTooltipChanged();
+    if (!!this.popupModel && !!this.popupModel.contentComponentData && !!this.popupModel.contentComponentData.model) {
+      const model = this.popupModel.contentComponentData.model;
+      if(Array.isArray(model.actions)) {
+        const actions: Array<any> = model.actions;
+        actions.forEach(item => {
+          if (!!(<any>item).locStrsChanged) {
+            (<any>item).locStrsChanged();
+          }
+        });
+      }
+    }
   }
   private locTitleChanged = () => {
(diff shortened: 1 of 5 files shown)


src/actions/action.ts
```diff
@@ -224,10 +233,35 @@ export class Action extends Base implements IAction, ILocalizableOwner {
...
+  public get locTitleName(): string {
+    return this.locTitle.localizationName;
+  }
+  public set locTitleName(val: string) {
+    this.locTitle.localizationName = val;
+  }
+  public locStrsChanged(): void {
+    super.locStrsChanged();
+    this.locTooltipChanged();
+    if (!!this.popupModel && !!this.popupModel.contentComponentData && !!this.popupModel.contentComponentData.model) {
```
tsv2013, line 245: This probably is better to move to ActionDropdownViewModel class owns popupModel

tsv2013 (changes requested):

src/actions/action.ts
andrewtelnov: Unfortunately it is a wrapper on action and it doesn't have locStrsChanged.
`export class ActionDropdownViewModel {`
...
```js
  constructor(private item: Action) {
    this.setupPopupCallbacks();
  }
```
I don't like the code myself. However, it is better to have it here than in Creator.

tsv2013 (approved):

PR #4904, opened 2022-09-26
A11y issues

attalvishal:
fixed eslint errors


andrewtelnov:
@attalvishal You have just changed the formating. There is no any code changed related to A11.

PR #4938, opened 2022-09-30
Update descriptions in the `QuestionSelectBase` and `QuestionCheckboxBase` classes

RomanTsukanov:
+

### DIFF
--- a/src/question_baseselect.ts
+++ b/src/question_baseselect.ts
@@ -14,7 +14,7 @@ import { settings } from "./settings";
 import { CssClassBuilder } from "./utils/cssClassBuilder";
 
 /**
- * It is a base class for checkbox, dropdown and radiogroup questions.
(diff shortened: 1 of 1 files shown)


src/question_baseselect.ts
```diff
@@ -485,7 +504,18 @@ export class QuestionSelectBase extends Question {
...
   /**
-   * The list of items. Every item has value and text. If text is empty, the value is rendered. The item text supports markdown.
+   * Gets or sets choice items. This property accepts an array of objects with the following structure:
+   *
+   * ```js
+   * {
+   *   "value": any, // A value to be saved in the survey results
+   *   "text": String, // A display text. This property supports Markdown. When `text` is undefined, `value` is used.
+   *   "imageLink": String // A link to the image or video that represents this choice value. Applies only to Image Picker questions.
+   *   "customProperty": any // Any property that you find useful
```
andrewtelnov, line 514: I would add an example how to add a new property, because if the property is not added into "itemvalue" our Serializer will not set it into object from JSON.

andrewtelnov (approved):

PR #5037, opened 2022-10-14
Skip randomizing matrix rows in design mode

jdpetit:
Matrix rows are randomized in the survey creator if rowsOrder is random, making management of the survey more difficult.


andrewtelnov:
@jdpetit Could catch. We need to update a unit test accordingly.

Thank you,
Andrew

andrewtelnov:
@jdpetit Thank you for your PR!

Andrwe

andrewtelnov (approved):

PR #5124, opened 2022-10-27
Fixed #5092 - RTL - A question number appears on the left side instead of the right side of a question title

### LINKED ISSUE #5092: RTL - A question number appears on the left side instead of the right side of a question title
A question number should appear on the right side in RTL mode.
[Example](https://plnkr.co/edit/e0zv5Zc7Lyo8Ut88)
image

tsv2013:


novikov82 (commented):
Is it possible to write a simple screenshot test?

novikov82 (approved):

PR #5142, opened 2022-10-31
Update the Ranking, Radiogroup, and Multiple Text doccomments

RomanTsukanov:


andrewtelnov (approved):

src/question_multipletext.ts
```diff
@@ -414,17 +420,14 @@ export class QuestionMultipleTextModel extends Question
     this.setPropertyValue("colCount", val);
   }
   /**
-   * The default text input size.
+   * A value passed on to the [`size`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size) attribute of the undelying `<input>` elements.
```
annnke, line 423:
```suggestion
   * A value passed on to the [`size`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size) attribute of the underlying `<input>` elements.
```

annnke (approved):

PR #5145, opened 2022-10-31
Update the Signature Pad and Text doccomments

RomanTsukanov:


andrewtelnov (approved):

src/question_signaturepad.ts
```diff
@@ -172,7 +181,8 @@ export class QuestionSignaturePadModel extends Question {
     return !this.isInputReadOnly && this.allowClear;
   }
   /**
-   * Use it to set pen color for the signature pad.
+   * Speicifies a color for the pen. Accepts hexadecimal colors (`"#FF0000"`), RGB colors (`"rgb(255,0,0)"`), or color names (`"red"`).
```
annnke, line 184:
```suggestion
   * Specifies a color for the pen. Accepts hexadecimal colors (`"#FF0000"`), RGB colors (`"rgb(255,0,0)"`), or color names (`"red"`).
```

src/question_text.ts
```diff
@@ -343,7 +349,7 @@ export class QuestionTextModel extends QuestionTextBase {
     return ["date", "datetime", "datetime-local"].indexOf(this.inputType) < 0;
   }
   /**
-   * The list of recommended options available to choose.
+   * An array of predefined options that from which users can select. This property configures an HTML [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) element and associates it with the underlying `input` element.
```
annnke, line 352:
```suggestion
   * An array of predefined options from which users can select. This property configures an HTML [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) element and associates it with the underlying `input` element.
```

annnke (approved):

PR #5223, opened 2022-11-11
PR: Accessibility Issue for a Required Checkbox: Ensures every ARIA input field has an accessible name

dmitrykurmanov:
work for the [#3649](https://github.com/surveyjs/survey-creator/issues/3649)


novikov82 (commented):
What about markup tests for new changes?

tsv2013 (commented):
Can we add unit or markup tests here?

tsv2013 (approved):

Issue #5267, opened 2022-11-21
Timer design issues


dk981234:
- [x] fix timer shadow
- [x] implement support for clock timer with no limits
- [ ] fix quiz example (change timer and progress location: timer -> "bottom", "progress" -> "top")

PR #5271, opened 2022-11-22
Work for #5267: do not show clock if limits are not set

dk981234:


andrewtelnov:
@dk981234 Could we show clocks, but only show the number of seconds spent in survey, from 0 to unlimited number?

tsv2013 (approved):

PR #5333, opened 2022-12-13
Updated .sd-timer__progress | stroke

JaneSjs:


dk981234 (approved):

src/defaultV2-theme/blocks/sd-timer.scss
```diff
@@ -31,7 +31,7 @@
   height: 100%;
   width: 100%;
   transform: rotate(-90deg);
-  stroke: #19B394;
+  stroke: $primary;
```
OlgaLarina, line 34: Maybe it's better to specify the color like this?
`var(--primary, #f3f3f3);`

JaneSjs: @OlgaLarina, we already specify ` #19B394` as an optional second argument for the `primary` variable at https://github.com/surveyjs/survey-library/blob/master/src/defaultV2-theme/variables.scss#L1. So I believe this is unnecessary.

dk981234 (approved):

PR #5394, opened 2022-12-21
fix(matrix): export IMatrixDropdownData to allow extending MatrixDynamicRowModel

maxpowa:
Since this is not currently exported, type checks will fails if you try to keep types consistent with the internal surveyjs types when extending MatrixDynamicRowModel.


tsv2013 (approved):

tsv2013:
@maxpowa Thank you for the contribution!

PR #5710, opened 2023-02-24
Implement design for dropdown in tablet mode

dk981234:


novikov82 (approved):

novikov82 (changes requested):
I see the "magic number" 600 in this method. Is there any ways to avoid this unnamed constant?
```
 private calculateIsTablet (windowWidth: number, windowHeight: number) {
    const width = Math.min(windowWidth, windowHeight);
    this.isTablet = width >= 600;
  }
```

novikov82 (approved):

PR #5730, opened 2023-03-02
Lazy Loading - A default value is not saved to survey results

OlgaLarina:


andrewtelnov (commented):
Will we have the same issue for tagbox?

andrewtelnov (approved):

PR #6318, opened 2023-06-06
Property Grid Choices Editor - Allow drag items using the drag button rather than the entire item area  

dmitrykurmanov:
see https://github.com/surveyjs/survey-creator/issues/4174


src/survey.ts
```diff
@@ -7065,6 +7079,10 @@ Serializer.addClass("survey", [
     default: "initial",
     choices: ["initial", "random"],
   },
+  {
+    name: "matrixDragHandleArea",
+    default: "entireItem"
```
tsv2013, line 7084: choices: ["entireItem", "icon"]

src/base-interfaces.ts
```diff
@@ -55,6 +55,7 @@ export interface ISurvey extends ITextProcessor, ISurveyErrorOwner {
   isClearValueOnHidden: boolean;
   isClearValueOnHiddenContainer: boolean;
   questionsOrder: string;
+  matrixDragHandleArea: string;
```
tsv2013, line 58: Why do we need to add this property to this interface?

Issue #6478, opened 2023-07-05
Do not use JSON stringify/parse functions in Helpers.getUnbindValue


andrewtelnov:
Using JSON stringify/parse function can lead to issues if using it for a array with large strings (files content in base64).
[Here is the related issue on SurveyJS support forum](https://surveyjs.answerdesk.io/ticket/details/T13659).

PR #6479, opened 2023-07-05
Do not use JSON stringify/parse functions in Helpers.getUnbindValue f…

andrewtelnov:
…ix #6478

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -162,13 +162,27 @@ export class Helpers {
     return array;
   }
   public static getUnbindValue(value: any): any {
+    if(Array.isArray(value)) {
+      const res = [];
+      for(let i = 0; i < value.length; i ++) {
+        res.push(Helpers.getUnbindValue(value[i]));
+      }
+      return res;
+    }
     if (!!value && value instanceof Object && !(value instanceof Date)) {
-      //do not return the same object instance!!!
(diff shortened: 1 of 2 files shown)


tsv2013 (approved):

src/helpers.ts
```diff
@@ -162,13 +162,27 @@ export class Helpers {
     return array;
   }
   public static getUnbindValue(value: any): any {
```
salargl, line 164: Hey @andrewtelnov thanks for the fix. But I wonder what this function tries to achieve?

andrewtelnov: @salargl  If you run unit tests with your code then 68 unit tests will be failed.
Here is the reason:
```js
const val = matrixQustion.value;
val.push(newValue); //this line will modify matrixQustion.value becaue the instance is the same.
```
I had to modify the code to the following to fix unit tests in SurveyJS Creator V1&V2:
```js
  public static getUnbindValue(value: any): any {
    if(Array.isArray(value)) {
      const res = [];
      for(let i = 0; i < value.length; i ++) {
        res.push(Helpers.getUnbindValue(value[i]));
      }
      return res;
    }
    if (!!value && value instanceof Object && !(value instanceof Date)) {
      return JSON.parse(JSON.stringify(value));
    }
    return value;
  }

```

Thank you,
Andrew

PR #6514, opened 2023-07-11
fix(popup): prevent exception on Legacy Edge when opening a dropdown popup

maxpowa:
Legacy Edge does not support the `VisualViewport` browser API. This causes an exception in the popup logic and breaks the UI in an unrecoverable manner. Legacy Edge also happens to be the engine which powers WinUI WebViews, which react-native-windows uses by default.


maxpowa:
I am not sure why the file decided to reformat as well... Let me revert the unrelated changes.

tsv2013 (approved):

Issue #6525, opened 2023-07-13
Let dropdown search-on-type match characters ignoring typographical marks (A when using Á, Ä...)


PerikiyoXD:
### Are you requesting a feature, reporting a bug or asking a question?
A feature

### What is the current behavior?
It will perfect-match the string

### What is the expected behavior?
It will match a non typographic mark version of the string
'Álava' -> 'Alava'

### How would you reproduce the current behavior (if this is a bug)?
N/A

#### Provide the test code and the tested page URL (if applicable)

Tested page URL:
https://surveyjs.io/create-free-survey

Test code
```
{
 "locale": "es",
 "title": {
  "es": "typographic marks test"
 },
 "pages": [
  {
   "name": "page1",
   "elements": [
    {
     "type": "dropdown",
     "name": "question1",
     "choices": [
      {
       "value": "Item 2",
       "text": {
        "es": "Álava"
       }
      },
      {
       "value": "Item 3",
       "text": {
        "es": "Mérida"
       }
      },
      {
       "value": "Item 4",
       "text": {
        "es": "Cáceres"
       }
      }
     ]
    }
   ]
  }
 ],
 "showQuestionNumbers": "off"
}
```

### Specify your

* browser: Firefox
* browser version: 114.0.2 (64-bit)
* surveyjs platform (angular or react or jquery or knockout or vue): jquery
* surveyjs version: v1.9.97

PerikiyoXD:
Any feedback is much appreciated

PerikiyoXD:
I've made some simple JS function that does what I want

```js
function removeDiacritics(text) {
  if (typeof text !== 'string') {
    throw new Error('Input should be a string.');
  }
  
  return text.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
}

console.log(removeDiacritics('Ávila')); // Output: Avila
console.log(removeDiacritics('Cáceres')); // Output: Caceres
console.log(removeDiacritics('Tórrido del Zujar')); // Output: Torrido del Zujar

```

PerikiyoXD:
I hope we can make this into a PR with proper code placement, this impacts latin-derived diacritic-featured languages. Thank you for your time!

andrewtelnov:
@PerikiyoXD Did you make a PR? We are going to discuss it on our sprint planning next week.

Thank you,
Andrew

PerikiyoXD:
[...]

PR #6572, opened 2023-07-23
fix(#6525):Add string normalization to hasText method

PerikiyoXD:
This is a fix for #6525

### DIFF
--- a/src/list.ts
+++ b/src/list.ts
@@ -64,7 +64,7 @@ export class ListModel<T extends BaseAction = Action> extends ActionContainer<T>
 
   private hasText(item: T, filterStringInLow: string): boolean {
     if (!filterStringInLow) return true;
-    let textInLow = (item.title || "").toLocaleLowerCase();
+    let textInLow = (item.title || "").toLocaleLowerCase().normalize("NFD").replace(/[\u0300-\u036f]/g, "");
     return textInLow.indexOf(filterStringInLow.toLocaleLowerCase()) > -1;
   }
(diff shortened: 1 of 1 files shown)


tsv2013 (commented):
A test (unit or functional) is nesessary. Otherwise functionality can be lost during future work.

PerikiyoXD:
I'm not familiar with QUnit, I'll investigate and try do some unit testing

PerikiyoXD:
I need help, I'd need to know where to place the test file once I got it ready.

andrewtelnov:
@PerikiyoXD Ι can add a unit test, However, I would make it optional, If you don't mind.
I would add a new option [here](https://github.com/surveyjs/survey-library/blob/master/src/settings.ts#L450) and make it `false` by default.

Thank you,
Andrew

PerikiyoXD:
I'll have a look at it today.

If you can add the test, please do.
 
The configuration will be named `diacriticSensitive`, and will be `false` by default.

I guess this is enough for the test to be made.

PerikiyoXD:
I'm okay with tests being labeled as "optional," but it's essential to include these optional tests in the CI/CD process as well.
Otherwise, if a change accidentally removes this optional status (which ideally should not happen), it won't be caught.
I'm not familiar with the specific setup of the CI/CD pipeline or whether you schedule all test runs, but I just wanted to bring this to your attention.

tsv2013:
Closing in favor of @andrewtelnov PR

PR #6746, opened 2023-08-17
remove unneeded paddings in row title cell

novikov82:
https://github.com/surveyjs/survey-creator/issues/4431


novikov82:
reverted

PR #6778, opened 2023-08-23
PR: Survey Creator landing: Drag and drop doesn't work

dmitrykurmanov:
the `hidden` prop doesn't work with `flex` containers https://stackoverflow.com/questions/23772673/hidden-property-does-not-work-with-flex-box

see  https://github.com/surveyjs/service/issues/1346


novikov82 (commented):
What about tests?

novikov82 (approved):

PR #6907, opened 2023-09-07
#6877 Mobile: Do not use min size for mobiles

### LINKED ISSUE #6877: Mobile: Do not use min size for mobiles
We should ignore min size for mobiles, since we have one question per row.

image

novikov82:
Fixes #6877


src/question.ts
```diff
@@ -101,6 +101,7 @@ export class Question extends SurveyElement<Question>
   protected setIsMobile(val: boolean) { }
 
   @property({ defaultValue: false, onSet: (val: boolean, target: Question) => {
+    target.renderMinWidth = !val;
```
andrewtelnov, line 104: Do we have renderMinWidth property?

novikov82: Yep, I added it to SurveyElement class 
https://github.com/surveyjs/survey-library/pull/6907/files#diff-d042e63e655cfdcd2783a19d550645690de316e2e69aca982f9b52c30d5ad393R910

andrewtelnov (approved):

PR #6930, opened 2023-09-10
Make sure we do not dispose object twice. fix #6929

### LINKED ISSUE #6929: Make sure we do not dispose object twice.
We need to move the code that clean objects into disposeCore and `dispose` should make checks if the object is already disposed or not.

andrewtelnov:


tsv2013 (commented):
We definitely shouldn't call dispose twice. But I believe if we call dispose twice it means we have wrong dispose order or some "live" still mistakenly holds a reference to a "dead" instance. I thought the isDisposed flag is not an option. We need to discuss it.

Issue #6859, opened 2023-09-04
VoiceOver doesn't read question titles because `aria-label` is missing from question title HTML elements


JaneSjs:
T14457 - Re: Accessibility Issue: Aria label missing.
https://surveyjs.answerdesk.io/internal/ticket/details/T14457

related issues:

- https://github.com/surveyjs/survey-library/issues/4121
- https://github.com/surveyjs/survey-library/issues/3421
- https://github.com/surveyjs/survey-library/pull/6210

dmitrykurmanov:
https://stackoverflow.com/questions/42477371/voiceover-not-reading-aria-label-for-links-in-ios-10
https://bugs.webkit.org/show_bug.cgi?id=162578

JaneSjs:
Upon further research, it appears that the title of a Radiogroup question is not being read by a Voice Over screen reader when the Radiogroup appears on the first survey page and the [survey.focusFirstQuestionAutomatic](https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#focusFirstQuestionAutomatic) option is enabled (which is the default setting).

However, the structure of the resulting form page seems to comply with accessibility standards. Additionally, we have tested the survey with other screen readers and confirmed that the title of a radiogroup question is correctly read by them (for example, we tested it with NVDA). Therefore, it seems to be a specific issue that is only reproduced on the Voice Over reader. Unfortunately, we may not be able to address this issue at the moment.

To resolve this issue, we recommend disabling the [survey.focusFirstQuestionAutomatic](https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#focusFirstQuestionAutomatic) option. With this option disabled, Voice Over should correctly read the title of a radiogroup question even if the question appears on the first survey page.

PR #6937, opened 2023-09-11
PR: A11Y:  Radiogroup and Checkbox (voice over + better structure)

dmitrykurmanov:
see https://github.com/surveyjs/survey-library/issues/6859


testCafe/questions/radiogroup.js
```diff
@@ -74,13 +74,13 @@ frameworks.forEach(framework => {
 
     //asc
     await setOptions("car", { choicesOrder: "asc" });
-    await t.expect(chocies.nth(0).find("[aria-label=\"Audi\"]").exists).ok();
-    await t.expect(chocies.nth(3).find("[aria-label=\"BMW\"]").exists).ok();
+    await t.expect(chocies.nth(0).find("[type=\"radio\"]").exists).ok();
```
tsv2013, line 77: It looks like here we checked items order using aria label. Changes Audi/BMW to radio/radio break the test. Probably other checks should be also validated.

packages/survey-angular-ui/src/comment-other.component.html
```diff
@@ -1,4 +1,4 @@
-<textarea *ngIf="!question.isReadOnlyRenderDiv()" [id]="otherId" [attr.maxlength]="question.getOthersMaxLength()"  [attr.aria-required]="question.ariaRequired" [attr.aria-label]="question.ariaLabel" [attr.placeholder]="otherPlaceholder" 
+<textarea *ngIf="!question.isReadOnlyRenderDiv()" [id]="otherId" [attr.maxlength]="question.getOthersMaxLength()"  [attr.aria-required]="question.ariaRequired || question.a11y_input_ariaRequired" [attr.aria-label]="question.ariaLabel || question.a11y_input_ariaLabel" [attr.placeholder]="otherPlaceholder"
```
tsv2013, line 1: Why do we have two checks - question.ariaRequired || question.a11y_input_ariaRequired
The quesion is required or not required. Why a11y_input_ariaRequired could be different from the ariaRequired?

packages/survey-vue3-ui/src/QuestionComment.vue
```diff
@@ -11,8 +11,8 @@
     :value="question.comment"
     :maxlength="question.getOthersMaxLength()"
     :placeholder="question.renderedCommentPlaceholder"
-    :aria-label="question.ariaLabel"
-    :aria-required="question.ariaRequired"
+    :aria-label="question.ariaLabel || question.a11y_input_ariaLabel"
```
tsv2013, line 14: question.ariaLabel || question.a11y_input_ariaLabel
It looks like they always should be the same.

src/question.ts
```diff
@@ -2335,6 +2313,78 @@ export class Question extends SurveyElement<Question>
       this.dependedQuestions[i].resetDependedQuestion();
     }
   }
+
```
tsv2013, line 2316: Do we have a unit test for these calculations?

tsv2013 (approved):

PR #7120, opened 2023-10-11
Fix theme event declaration

RomanTsukanov:


RomanTsukanov:
This event is going to be deleted.

PR #7175, opened 2023-10-18
add headerRoot & bodyRoot to page

ryan-feagin:
wrap the page header and body in div with classNames to allow for more css control over layout.  Allows for sticky page header or to use flex on large screens to 'split' the screen with page header to the left and questions to the right.


tsv2013:
@ryan-feagin We highly appreciate your contribution in our project. At this moment surveyjs library is available for the following frameworks: react, knockout, angular, vue and vue3. And all the markup and CSS classes are identical for all the supported platforms. That's why for PR realated to markup changes it is necerssary to implement similar changes for all the mentioned platforms and fix new markup in markup or visual tests.
Thank you for the understanding.

Issue #7219, opened 2023-10-24
Signature Pad is not Responsive on Mobile Screens


JaneSjs:
[View Demo](https://codesandbox.io/s/dry-smoke-6d99qj?file=/src/json.js)

If a Signature Pad has a large signature width, the signature area is not responsively adjusted. On Mobile screens, the following issue occurs:
![image](https://github.com/surveyjs/survey-library/assets/22372972/13dc37aa-2f1c-4a28-a24d-daf72e2fb837)

PR #7291, opened 2023-11-04
#7219 - fix signature scaling

novikov82:
Critical fix before release!


dk981234 (approved):

CodeCurosity:
using the latest file for jquery but width & height is not working properly?

PR #7367, opened 2023-11-16
PR (LIBRARY): Imagepicker item looses controls after dragging

dmitrykurmanov:
functional test is here https://github.com/surveyjs/survey-creator/pull/4900
see https://github.com/surveyjs/survey-creator/issues/4888


dk981234 (commented):
Maybe we should use class instead of inline styles for imagepicker item's  drag shortcut

dmitrykurmanov:
> Maybe we should use class instead of inline styles for imagepicker item's drag shortcut

will be fixed in https://github.com/surveyjs/survey-library/issues/7368

dk981234 (approved):

PR #7503, opened 2023-12-14
Fixed https://github.com/surveyjs/private-tasks/issues/342 - Survey Header + Progress Bar

tsv2013:


src/survey.ts
```diff
@@ -7390,6 +7402,7 @@ export class SurveyModel extends SurveyElementCore
           }
         }
       } else if (isStrCiEqual(layoutElement.id, "advanced-header")) {
+        const header = layoutElement.data as Cover;
```
novikov82, line 7405: const header is not used in this scope. maybe we can delete this line?

tsv2013: Forgot it. Removed

novikov82 (approved):

PR #7504, opened 2023-12-15
Work for #7502: implement fadeIn/fadeOut animation for popup

dk981234:


testCafe/survey/popup.js
dmitrykurmanov, line 1: probably we don't need it because of helper.js setting

tests/components/popuptests.ts
dmitrykurmanov, line 1: it seems that it would be better to use `Survey.settings.animationEnabled` here instead of using `getShouldRunAnimation` directly

tests/markup/etalon_dropdown.ts
dmitrykurmanov, line 1: I think it would  be helpful to disable animation for all markup and functional tests

src/utils/devices.ts
dmitrykurmanov, line 1: ^_^

src/utils/animation.ts
dmitrykurmanov, line 1: do we need something like that?

```js
animationFrameId = requestAnimationFrame(...);
cancelAnimationFrame(animationFrameId);
```
and `clearTimeout()` for  setTimeout

tsv2013 (approved):

PR #7517, opened 2023-12-18
Fix: Prevent clear of other questions when question name equals other value name

marvin-wtt:
By default, the values of all invisible questions are cleared on complete to exclude the data from the result. As multiple questions may have the same value name, it is checked wherever there is another question with the same value name is visible before clearing the value. 
However, this check does not take into account, that the question name is the default value name when no value name is provided. This causes all questions with the same value name as the question name to be cleared.

This PR suggests using `getValueName` instead, which falls back to the question name if no value name is defined.

<details>
  <summary>Example Survey</summary>
  
```json
{
 "pages": [
  {
   "name": "page1",
   "elements": [
    {
     "type": "boolean",
     "name": "toggle",
     "title": "Toggle",
     "labelTrue": "Bug",
     "labelFalse": "It works"
    },
    {
     "type": "text",
     "name": "username",
     "visibleIf": "{toggle} = false",
     "title": "Username",
     "maxLength": 25
    },
    {
     "type": "text",
     "name": "alternative_username",
     "visibleIf": "{toggle} = true",
     "title": "Alternative Username",
     "valueName": "username"
    }
   ]
  }
 ],
 "showQuestionNumbers": "off"
}
```
</details>

### DIFF
--- a/src/question.ts
+++ b/src/question.ts
@@ -1549,8 +1549,7 @@ export class Question extends SurveyElement<Question>
     if (this.isVisibleInSurvey) return false;
     if (!!this.page && this.page.isStartPage) return false;
     if (!this.survey) return true;
-    const valueName = this.valueName ?? this.name;
-    return !this.survey.hasVisibleQuestionByValueName(valueName);
+    return !this.survey.hasVisibleQuestionByValueName(this.getValueName());
   }
   /**
    * Returns `true` if a parent element (page or panel) is visible.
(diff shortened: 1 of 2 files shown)


andrewtelnov:
@marvin-wtt We need a unit test for this case to make sure we don't have the regression on a code moification. We don't accept PR without a unit test or functional test or visual regression test.
I can add it if you can't add it.

Thank you,
Andrew

marvin-wtt:
@andrewtelnov I can try to add a test. I noticed that your other tests usually link to an issue describing the problem. Should I create an issue for this too?

andrewtelnov:
@marvin-wtt Adding a unit test without creating issue is fine.

Thank you,
Andrew

PR #7629, opened 2024-01-05
fix: 🐛 resolved globalThis is undefinded in react-native-window

shupingchu:
globleThis is showing `undefinded` in react-native-windows Edge browser, this simple solution is to check the type if is `undefinded`, if so, use `self` instead, which is working in Edge

My Windows environment is:
react-native-windows: 0.69.6
survey-core: 1.9.121
survey-react-ui: 1.9.121,
Microsoft Edge DevTools debugger - Edge browser

## Screenshot
Microsoft Edge DevTools Preview 1_4_2024 4_15_12 PM

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -11,7 +11,7 @@ export type ISurveyEnvironment = {
   svgMountContainer: HTMLElement | string,
   stylesSheetsMountContainer: HTMLElement,
 }
-const document = globalThis.document;
+const document = typeof globalThis !== "undefined" ? globalThis.document : (this as any).document;
 const defaultEnvironment: ISurveyEnvironment = <ISurveyEnvironment>(!!document ? {
   root: document,
 


src/settings.ts
```diff
@@ -11,7 +11,7 @@ export type ISurveyEnvironment = {
   svgMountContainer: HTMLElement | string,
   stylesSheetsMountContainer: HTMLElement,
 }
-const document = globalThis.document;
+const document = typeof globalThis !== "undefined" ? globalThis.document : self.document;
```
tsv2013, line 14: Probably we need to change to the `this.document;`

tsv2013 (changes requested):

src/settings.ts
shupingchu: Since `this.document` will failed on eslint check due to the possible undefined, I need to do cast as any
```
[6] ERROR in C:\Users\janni\Downloads\survey-library\src\settings.ts
[6] [tsl] ERROR in C:\Users\janni\Downloads\survey-library\src\settings.ts(14,18)
[6]       TS2532: Object is possibly 'undefined'.

Comment (tsv2013) [reply src/settings.ts]: I meant
```
const document = typeof globalThis !== "undefined" ? globalThis.document : this.document;
```

tsv2013: I'm afraid `self` can be undefined under some circumstances

tsv2013 (approved):

tsv2013:
@shupingchu Thank you for the contribution!

PR #7822, opened 2024-02-07
Multi-Select/Dynamic Matrix: Deprecate columnsLayout, implement the t…

### LINKED ISSUE #7807: Multi-Select/Dynamic Matrix: Deprecate `columnsLayout`, implement the `transposeData: true | false` property instead

andrewtelnov:
…ransposeData: true | false property instead fix #7807

### DIFF
--- a/src/question_matrixdropdownbase.ts
+++ b/src/question_matrixdropdownbase.ts
@@ -900,7 +900,7 @@ export class QuestionMatrixDropdownModelBase extends QuestionMatrixBaseModel<Mat
     );
     this.registerPropertyChangedHandlers(
       [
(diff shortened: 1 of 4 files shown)


src/question_matrixdropdownbase.ts
```diff
@@ -959,6 +959,12 @@ export class QuestionMatrixDropdownModelBase extends QuestionMatrixBaseModel<Mat
       this.clearRowsAndResetRenderedTable();
     }
   }
+  public get transposeData(): boolean {
```
novikov82, line 962: doc comments?

tsv2013 (approved):

PR #7865, opened 2024-02-14
Work for https://github.com/surveyjs/service/issues/1933 - A SurveyJS Creator flickers when switching tabs

tsv2013:


tsv2013:
I've found the actual reason. Closing in favor of https://github.com/surveyjs/survey-creator/pull/5230

PR #7926, opened 2024-03-05
Fix matrix's row text show twice in modern, default theme

dk981234:


tsv2013 (approved):

momaekwudi:
It will cause mobile rich text parsing to fail

PR #7938, opened 2024-03-08
Fix Spelling and Add Other Cases

JavaScriptDude:
My iPhone does not use user or environment in the device label. Instead they use Front and Back. Also environment was spelled incorrectly.


andrewtelnov (approved):

andrewtelnov:
@JavaScriptDude I have created a [new PR](https://github.com/surveyjs/survey-library/pull/9066) to fix unit tests & resolve conflics.

Thank you,
Andrew

PR #8032, opened 2024-03-29
Update hebrew.ts

UZ1sFED3yS:
Word used for "clearCaption" translation to hebrew is wrong in this context. 

Used translation is right in context of, for example, "clear sky", or "it's clear", but wrong in context of "to clear something'


tsv2013 (approved):

tsv2013:
Thank you for the contribution, @UZ1sFED3yS !

PR #8118, opened 2024-04-15
#8110 Matrix. Areas have different colors in some themes.

### LINKED ISSUE #8110: Matrix. Areas have different colors in some themes.
Image

novikov82:
Fixes #8110


tsv2013 (approved):
I'd like us to avoid long nested CSS rules usage. But most likel;y we'll be able to make it after matrix re-layout only.

PR #8681, opened 2024-08-13
Describe `validateVisitedEmptyFIelds`

RomanTsukanov:


packages/survey-core/src/survey.ts
```diff
@@ -1725,6 +1725,15 @@ export class SurveyModel extends SurveyElementCore
   public set checkErrorsMode(val: string) {
     this.setPropertyValue("checkErrorsMode", val);
   }
+  /**
```
andrewtelnov, line 1728: It works for text, comment, dropdown and tagbox (it doesn't work for tagbox right now)

andrewtelnov (approved):

PR #8734, opened 2024-08-23
Expression doesn't work if inside the sync function there is an async…

### LINKED ISSUE #8732: Expression doesn't work if inside the sync function there is an async function as a parameter
The folllowing expression doesn't work:
`syncFunction(asyncFunction({question1})`.

andrewtelnov:
… function as a parameter fix #8732


WuerzburgerMaetschke:
amazing, thx for the fast fix!

tsv2013 (approved):

Issue #8911, opened 2024-10-07
Remove usage of window global


kesavkolla:
It's better to use globalThis instead of window. This will align with the modern usage of global variables.

PR #8912, opened 2024-10-07
Replace window with globalThis

kesavkolla:
Replaced window with globalThis. This will fix the issue #8911

### DIFF
--- a/src/entries/js-ui.ts
+++ b/src/entries/js-ui.ts
@@ -65,7 +65,9 @@ function doPopupSurvey(props: any): void {
   });
 }
 
-let jQueryInst = globalThis["jQuery"] || globalThis["$"];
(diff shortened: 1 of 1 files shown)


dmitry-kurmanov:
@kesavkolla Thanks for the contribution! 

Unfortunately, according to https://caniuse.com/?search=globalThis the `globalThis` does not supported by IE11. Probably we could increase the PR via some extra code:

```js
let gcontext = globalThis;

if (typeof globalThis === "undefined") context = window;
 
let jQueryInst = context ["jQuery"] || context ["$"];
```

What do you think?

kesavkolla:
Added check for globalThis and fallback to window.

dmitry-kurmanov (approved):

PR #8917, opened 2024-10-09
fix - makes dropdown type of rating question display correct values on options

lander854:
Rating questions can be turned into dropdowns when the screen width is not enough to fit all answers. When using dropdowns, values were used instead of text for dropdown options.

### DIFF
--- a/packages/survey-core/src/question_rating.ts
+++ b/packages/survey-core/src/question_rating.ts
@@ -37,8 +37,8 @@ export class RenderedRatingItem extends Base {
 }
 
 class RatingItemValue extends ItemValue {
-  constructor(value: any, public description: LocalizableString) {
-    super(value);
+  constructor(value: any, text: string = null, public description: LocalizableString) {
+    super(value, text);
   }
 }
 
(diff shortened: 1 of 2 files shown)


tsv2013:
@lander854 Thank you for contributing and sorry for late response! But there are some moments blocking this PR:
- we're using tests - unit, functional or visual - to provide product stability
- we support 4 frameworks - ReactJS, AngularJS, VueJS 3 (and 2 for a while) and KnockoutJS - and all the functionality we implement should be the same for all these frameworks

If seems to getting too much of your time, you can create an issue here on GitHub and our team implements it as soon as task will be added to sprint.

tsv2013 (changes requested):
There are some moments blocking this PR:

we're using tests - unit, functional or visual - to provide product stability
we support 4 frameworks - ReactJS, AngularJS, VueJS 3 (and 2 for a while) and KnockoutJS - and all the functionality we implement should be the same for all these frameworks

novikov82:
We created another PR https://github.com/surveyjs/survey-library/pull/8968 using your code changes. Unit tests are added.

PR #9017, opened 2024-11-07
Implement help methods for creator's d&d pages animation

dk981234:


packages/survey-core/src/page.ts
```diff
@@ -317,6 +317,14 @@ export class PageModel extends PanelModelBase implements IPage {
     super.ensureRowsVisibility();
     this.getPanels().forEach((panel) => panel.ensureRowsVisibility());
   }
+  public onDetachFromUICallback: () => void;
```
tsv2013, line 320: I don't like this callback. We need to discuss what we could do here.

tsv2013 (approved):

PR #9482, opened 2025-02-19
Clear incorrect values if needed on question validation in select base

andrewtelnov:


tylers:
I believe this change broke the ability to select options that were retrieved from external sources when applying the _select2_ `processResults() `method.

If you use _select2_ initialization with the _tagbox_ element, then fetch new options based on the query/term entered. Once you select an option, you can verify the question will have it's value updated with SurveyJS's `onValueChanged()` method. But it will immediately remove that option from the value and `onValueChanged()` will fire again. 

In this example:

- click on _tagbox_ question and it opens the dropdown panel
- search for term '99'
- new options are fetched and added to the panel
- select a new option
- `onValueChanged()` is updated with value ['99']
- `onValueChanged()` is updated with value []

I'm not sure if this is the change that caused it but the functionality broke between `1.12.23` and `1.12.24` and this is the ticket that seems the most likely to have caused it. I don't have a sandbox to demonstrate and I couldn't find an example that retrieves options with _select2_ on the site to extend.

@tsv2013

Issue #9487, opened 2025-02-20
Radio Button Group / Checkboxes: In multi-column layout, items are arranged unnaturally


RomanTsukanov:
![Image](https://github.com/user-attachments/assets/d54073a2-cfa7-4cc6-98e8-167b63040a13)

```js
{
  "title": "NPS Survey Question",
  "pages": [
    {
      "name": "page1",
      "elements": [
        {
          "type": "radiogroup",
          "name": "question1",
          "choices": [
            "Item 1",
            "Item 2",
            "Item 3",
            "Item 4",
            "Item 5",
            "Item 6",
            "Item 7",
            "Item 8",
            "Item 9",
            "Item 10"
          ],
          "colCount": 3
        }
      ]
    }
  ]
}
```

PR #9522, opened 2025-02-25
PR: Radio Button Group / Checkboxes: In multi-column layout, items are arranged unnaturally

dmitry-kurmanov:
see https://github.com/surveyjs/survey-library/issues/9487


packages/survey-core/src/question_baseselect.ts
```diff
@@ -1931,40 +1931,48 @@ export class QuestionSelectBase extends Question {
...
+        if (choice) {
+          column.push(choice);
         }
       }
+      columns.push(column);
+    }
+
+    return columns;
+  }
+  getColumnsWithRowItemFlow(choices, colCount) {
```
tsv2013, line 1968: If we mark this method as private, we'll be free to rename/remove it in the future.

tsv2013 (approved):

Issue #9374, opened 2025-01-30
Specialized question type + showOtherItem doesn't show stored comment value


c-meier:
**Describe the bug**
When using a specialized question type that set `showOtherItem` to true, the comment box does not show the stored value.

**Steps to reproduce**

```js
// Create specialized question type with other item
Survey.ComponentCollection.Instance.add({ name: "custom-dropdown", title: "My dropdown", inheritBaseProps: true,
  questionJSON: {
    type: "dropdown",
    choices: ["Item 1", "Item 2", "Item 3"],
    showOtherItem: true
  },
});
// Create survey with the same question once as standard and once as a specialized question type
const surveyJson = {
  elements: [{ name: "direct",
               type: "dropdown",
               choices: ["Item 1", "Item 2", "Item 3"],
               showOtherItem: true },
             { name: "specialized",
               type: "custom-dropdown" }]
};
const survey = new Survey.Model(surveyJson);
// Fill data as if from local storage
survey.data = { "direct": "other", "direct-Comment": "Hello",
                "specialized": "other", "specialized-Comment": "World" };
function alertResults(sender) {
  const results = JSON.stringify(sender.data);
  alert(results);
}
survey.onComplete.add(alertResults);
document.addEventListener("DOMContentLoaded", function () {
 survey.render(document.getElementById("surveyContainer"));
});
```

**Expected behavior**

The comment box of the specialized question type should show it's comment value.

**Screenshots**

![Image](https://github.com/user-attachments/assets/feea3b4d-6fb5-4fca-89f2-7b8aa290db7c)

**Please complete the following information:**
- Browser: Firefox
- Browser version: 133.0.3
- JS framework/library: Vanilla JS
- SurveyJS version: 1.12.21
- Device: PC

JaneSjs:
Hi,
We researched a similar issue at https://github.com/surveyjs/survey-library/issues/9378. The fix will be available with the next maintenance update. Once a new version is available, please check to see if it works as expected.

Thanks

c-meier:
Hi,

I tested with https://unpkg.com/survey-core@2.0.0-rc.2/survey.core.min.js (which if the release notes are correct contains the fix) and it doesn't work.

c-meier:
Hi @JaneSjs,

I tested with 2.0.0-rc.6, it still doesn't work.

c-meier:
Fixed by version 2.1.0

PR #9588, opened 2025-03-11
Inelegant fix to: Specialized question type + showOtherItem doesn't show stored comment value #9374

c-meier:
I identified the source of the issue #9374, but I haven't found an elegant solution.

The issue is that `q.requireUpdateCommentValue` does not return the correct value, which prevents `q.updateCommentFromSurvey` from being called within `SurveyModel.updateAllQuestionsValue`.

This is a workaround that solves the problem, although it might not be the most elegant and could potentially cause issues in the future.

### DIFF
--- a/packages/survey-core/src/question_custom.ts
+++ b/packages/survey-core/src/question_custom.ts
@@ -842,6 +842,9 @@ export class QuestionCustomModel extends QuestionCustomModelBase {
       this.setValue(this.name, this.value, false, this.allowNotifyValueChanged);
     }
   }
+  public supportOther(): boolean {
+    return true;
+  }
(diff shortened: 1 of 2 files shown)


andrewtelnov:
@c-meier Thank you for pointing to the issue. It is an issue indeed. Let me think about the fix. I think we should have another fix for this. I will be back.

Thank you,
Andrew

andrewtelnov:
@c-meier Added comments.

Thank you,
Andrew

c-meier:
Fixed by version 2.1.0 (Commit 06d4ac6)

EDIT: may be interesting to add the tests (I didn't see any equivalent existing test with a quick look)

Issue #10190, opened 2025-07-31
JSON schema bug (surveyjs_definition.json)


nbelyh:
**Describe the bug**

The property `showPreviewBeforeComplete` allows boolean values (true/false). 
https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#showPreviewBeforeComplete

But in the surveyjs_definition.json file it only allows enums, therefore, you get a warning in vscode.

**Steps to reproduce**

1. configure schema for vscode
2. open any survey that has "showPreviewBeforeComplete": false
3. You see the warning above

**Expected behavior**
There should be no warning, the schema needs to allow boolean

**Screenshots**
<img width="549" height="308" alt="Image" src="https://github.com/user-attachments/assets/0316d3af-a844-4ddc-80b3-af736373090a" />

**Additional context**
I have created a PR to fix the issue.

JaneSjs:
Hi,
I tested the following JSON in my VSCode, however, didn't get any warnings.

<img width="1628" height="891" alt="Image" src="https://github.com/user-attachments/assets/e1f5b4c9-2927-49ac-97d1-0cb15e302a10" />

[schema.json](https://github.com/user-attachments/files/21543175/schema.json)

Would you please clarify how to reproduce the issue?

Thank you

nbelyh:
@JaneSjs 
You need to **enable schema support/validation** in VS Code for SurveyJS to see the issue. 
That's the only use of the schema files anyway, to enable validation/autocomplete😅

I have surveys files ending with "_survey", so I have in the settings.json file this:

<img width="786" height="194" alt="Image" src="https://github.com/user-attachments/assets/2b630394-9a52-4622-b337-c6298a9bb02f" />

settings.json
```jsonc
{
  "json.schemas": [
    {
      "fileMatch": [
        "/*_survey*.json" // pattern(s) to match your survey JSON files
      ],
      "url": "https://raw.githubusercontent.com/surveyjs/survey-library/refs/heads/master/packages/survey-core/docs/surveyjs_definition.json"
    },
  ],
```

nbelyh:
@JaneSjs 
I have attached a ready-to-reproduce example here (folder containing both a simple survey and an "enable schema" settings file:
Just unzip and open in VS Code (I've tried on my laptop). No extensions, just pure vs code.

[survey_schema.zip](https://github.com/user-attachments/files/21554330/survey_schema.zip)

[...]

PR #10191, opened 2025-07-31
Fix schema for showPreviewBeforeComplete

nbelyh:
Fix for the schema issue. Now the boolean is also allowed (it was allowed from some surveyjs version, but the schema was not reflecting this). Please see the related issue #10190

### DIFF
--- a/packages/survey-core/docs/surveyjs_definition.json
+++ b/packages/survey-core/docs/surveyjs_definition.json
@@ -393,11 +393,16 @@
             "type": "boolean"
         },
         "showPreviewBeforeComplete": {
-            "type": "string",
+            "type": [
+                "string",
+                "boolean"
+            ],
             "enum": [
                 "noPreview",
                 "showAllQuestions",
-                "showAnsweredQuestions"
+                "showAnsweredQuestions",
+                true,
+                false
             ]
         },
         "showTimerPanel": {


andrewtelnov:
@nbelyh It is the auto-generated JSON. 
[Serializer.generateSchema()](https://github.com/surveyjs/survey-library/blob/master/packages/survey-core/src/jsonobject.ts#L1506C10-L1506C24).

Thank you,
Andrew

nbelyh:
@andrewtelnov 
I see. Thank you for clarifying. So that is the one that should be corrected to fix the issue (linked), correct? I'm still interested in fixing it, it's annoying 🙂

Or is it not possible to fix this as of now (please look at the linked issue)?
Properties like `showPreviewBeforeComplete` allow boolean value now (true/false), but the schema does not allow that; the property is defined as "string" (enumeration). The survey creator automatically converts strings now to this boolean value, actually 🤔

andrewtelnov:
@nbelyh We have boolean & string type for this property to support backward compatibility.
In new versions you can set boolean type only. Here is the code:
```ts
  public get showPreviewBeforeComplete(): boolean | any {
    return this.getPropertyValue("showPreviewBeforeComplete");
  }
  public set showPreviewBeforeComplete(val: boolean | any) {
    this.setShowPreviewBeforeComplete(val);
  }
  public setShowPreviewBeforeComplete(val: boolean | any) {
    if (val === undefined || val === "noPreview" || val === false) {
      this.setPropertyValue("showPreviewBeforeComplete", false);
    } else {
      this.setPropertyValue("showPreviewBeforeComplete", true);
      if (val === "showAllQuestions")this.previewMode = "allQuestions";
      if (val === "showAnsweredQuestions")this.previewMode = "answeredQuestions";
    }
  }

```
So the boolean is correct.

Thank you,
Andrew
SurveyJS Team

nbelyh:
Thank you @andrewtelnov I understand that. The issue is however the schema does not really match to the actual values supported, ie basically this is not a show stopper in any way but a minor thing (schema validation warning)

I have tried to correct the schema manually in the PR but it looks like this is a wrong way to fix it, as the schema is autogenerated.

Would be probably okay to change the schema to support new way (booleans) and underscore the deprecated strings. Now it is the opposite (booleans are reported as invalid), and this is basically the issue.

Please take a look at the linked issue.

PR #10629, opened 2025-11-23
Bug/10618 matrix clear invisible

andrewtelnov:


andrewtelnov:
The PR contains code from getState/setState PR

PR #10672, opened 2025-12-02
Add support for promises in custom functions fix #10671

### LINKED ISSUE #10671: Add support for promises in custom functions

MarchTokenda:


packages/survey-core/tests/question_expressiontests.ts
```diff
@@ -1,6 +1,7 @@
 import { SurveyModel } from "../src/survey";
 import { QuestionExpressionModel } from "../src/question_expression";
 import { FunctionFactory } from "../src/functionsfactory";
+import { set } from "lodash";
```
tsv2013: It looks like a redundant import. We need to remove it.

PR #10727, opened 2025-12-22
LIBRARY: SSR and Shadow-Dom compatibility

dmitry-kurmanov:
https://github.com/surveyjs/private-tasks/issues/512

### DIFF
--- a/packages/survey-core/src/expressions/expressions.ts
+++ b/packages/survey-core/src/expressions/expressions.ts
@@ -644,13 +644,15 @@ export class OperandMaker {
       if (!left.length) {
         left = left.toString();
         if (typeof right === "string" || right instanceof String) {
-          left = left.toUpperCase();
-          right = right.toUpperCase();
         }
       }
       if (typeof left === "string" || left instanceof String) {
         if (!right) return false;
         right = right.toString();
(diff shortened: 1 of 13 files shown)


packages/survey-core/src/survey-element.ts
```diff
@@ -281,9 +281,12 @@ export class SurveyElement<E = any> extends SurveyElementCore implements ISurvey
     return res;
   }
   private static focusElementCore(elementId: string, containerEl?: HTMLElement): boolean {
-    const { root } = settings.environment;
-    if (!root && !containerEl) return false;
-    const el: HTMLElement = containerEl ? containerEl.querySelector(`#${CSS.escape(elementId)}`) : root.getElementById(elementId);
+    if (!containerEl) return false;
+    const rootNode = containerEl.getRootNode();
```
tsv2013: We could get rid of similar duplicated code (here and below) if we declare the following utility function:
```JS
function getValidRootNode(element) {
    const rootNode = !!element && element.getRootNode();
    if (rootNode instanceof Document || rootNode instanceof ShadowRoot) {
      return rootNode;
    }
    return undefined;
}
   ```

tsv2013 (approved):

PR #10878, opened 2026-02-04
Do not trim comments - replace strings containing only whitespace wit…

andrewtelnov:
Fixes https://github.com/surveyjs/survey-library/issues/10885

### DIFF
--- a/e2e/questions/checkboxes.spec.ts
+++ b/e2e/questions/checkboxes.spec.ts
@@ -334,7 +334,7 @@ frameworks.forEach((framework) => {
 
       const surveyResult = await getSurveyResult(page);
       expect(surveyResult.car).toStrictEqual(["other"]);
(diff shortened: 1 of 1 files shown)


packages/survey-core/src/question.ts
```diff
@@ -2613,19 +2613,18 @@ export class Question extends SurveyElement<Question>
...
   public set comment(newValue: string) {
+    newValue = this.getTrimmedComment(newValue);
     if (!!newValue && newValue !== this.comment) {
-      const trimmedValue = newValue.toString().trim();
-      if (trimmedValue !== newValue) {
-        newValue = trimmedValue;
-        if (newValue === this.comment) {
-          this.setPropertyValueDirectly("comment", newValue);
-        }
+      if (newValue === this.comment) {
```
OlgaLarina: This condition will never be true because we checked above that these values ​​are not equal

OlgaLarina (approved):

PR #11363, opened 2026-06-01
formbox hover tokens

novikov82:


tsv2013 (commented):
Do we indeed have hardcoded constants in fallbacks?

OlgaLarina (approved):

novikov82:
> Do we indeed have hardcoded constants in fallbacks?

I created issue [Get rid of fallbacks in our CSS-code when referencing --sjs2- vars](https://github.com/surveyjs/survey-library/issues/11367)

Issue #8901, opened 2024-10-04
A progress bar doesn't highlight previous pages when programmatically changing the survey.currentPage/currentPageNo


JaneSjs:
T20125 - Progress bar passed class not being set when resuming
https://surveyjs.answerdesk.io/internal/ticket/details/T20125
***
Consider the 'Continue Incomplete Survey' usage scenario. A developer programmatically specify the current page number to start a survey from the last visited page. However, the progress bar doesn't hightlight pages which were not actually visited by user but which were filled with answers previously.
[View Demo](https://plnkr.co/edit/maylNknTZtv0DFqQ)

![image](https://github.com/user-attachments/assets/6faa71e5-aff0-4178-bd7d-ffa076e5e86d)

JaneSjs:
To set pages visited, enable the `page.passed` option for target pages.

dmitry-kurmanov:
closed in favor of https://github.com/surveyjs/survey-library/issues/11132

JaneSjs:
This issue was reported at https://github.com/surveyjs/survey-library/pull/11429.

A progress bar doesn't highlight previous pages in v2.5.28: [View Demo](https://plnkr.co/edit/AdmgKqk7mliiDGKJ).

PR #11429, opened 2026-06-12
fix: mark preceding pages as passed on survey resume

nakazonot:
## Problem

When resuming a survey by setting `currentPageNo` to a value greater than 0,
progress bar buttons for pages before the current position were not shown as
"passed" (completed).

`currentPageChanged()` only marked the directly preceding page (`oldValue`) as
`passed = true`. Pages before the restored position were never marked,
leaving their indicators in an unvisited state.

## Fix

In `currentPageChanged()`, after marking `oldValue` as passed, loop through all
visible pages whose index is less than `newValue`'s and ensure they are also
marked as `passed = true`.

During normal sequential navigation this is a no-op (those pages are already
`passed`). The change only takes effect when jumping — most notably during
survey resume.

## Verification

- Added a unit test in `surveyProgressButtonsTest.ts` covering the resume scenario
- All 4326 existing unit tests pass (`vitest run`)
- Manually verified in the browser with `progressBarType: "buttons"` and `currentPageNo = 2`

### DIFF
--- a/packages/survey-core/src/survey.ts
+++ b/packages/survey-core/src/survey.ts
@@ -3980,6 +3980,15 @@ export class SurveyModel extends SurveyElementCore
         oldValue.passed = true;
       }
     }
+    if (newValue) {
+      const newIndex = this.visiblePages.indexOf(newValue);
+      for (let i = 0; i < newIndex; i++) {
+        const page = this.visiblePages[i];
+        if (!page.passed) {
+          page.passed = true;
+        }
+      }
+    }
     if (this.isCurrentPageRendered === true) {
       this.isCurrentPageRendered = false;
     }
--- a/packages/survey-core/tests/surveyProgressButtonsTest.ts
+++ b/packages/survey-core/tests/surveyProgressButtonsTest.ts
@@ -93,6 +93,22 @@ describe("ProgressButtons", () => {
     expect(progress.getListElementCss(1), "1) Page 2 style is empty").toBe("");
     expect(progress.getListElementCss(2), "1) Page 3 style is non clickable").toBe(survey.css.progressButtonsListElementNonClickable);
   });
+  test("ProgressButtons pages before current are marked passed on survey resume", () => {
+    const json: any = {
+      "pages": [
+        { "name": "page1", "elements": [{ "type": "text", "name": "q1" }] },
+        { "name": "page2", "elements": [{ "type": "text", "name": "q2" }] },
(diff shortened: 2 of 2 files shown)


JaneSjs:
Hi @nakazonot,
Thank you for your inquiry. We'll investigate this issue at https://github.com/surveyjs/survey-library/issues/8901. We'll inform you once we have any news to share.
Please stay tuned.

dmitry-kurmanov:
@nakazonot thanks for your contribution! After investigation, we fixed the problem with a more complex fix that fully covers your case. I am going to  close this PR in favor of https://github.com/surveyjs/survey-library/pull/11456.

nakazonot:
Thanks for the update and the thorough fix!

nakazonot:
Hi @dmitry-kurmanov, I tested `survey-core@2.5.33` (with #11456), but my original case is **still not resolved**.

#11456 fixed the progress bar when `currentPage`/`currentPageNo` is changed programmatically during a session. But when a survey is **resumed** with saved data, the preceding pages are still **not** marked as `passed`, so the progress bar doesn't show them as completed.

**Steps to reproduce**

1. Create a survey with 3 pages.
2. Load saved data and set `currentPageNo = 2` (resume on page 3).
3. The progress bar shows pages 1–2 as not passed, but they should be marked as passed.

The simple change in this PR does fix this resume case on my side. Could you reopen this (or #8901) to cover it?

Thanks!

JaneSjs:
Hi @nakazonot,
Thank you for the update.
We supported this usage sceanavio via the [`uiState`](https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#uiState) property. 
For more details on how to save the intermediate survey response and UI state, refer to the following demo: [Continue an Incomplete Survey](https://surveyjs.io/form-library/examples/save-and-restore-user-responses-to-complete-survey/).

To achieve your goal, use the `uiState` to persist the current page number. Refer to the following demo: [View Plunker](https://plnkr.co/edit/VgIv23rlaBLy1V8e).
```js
const prevState = {
  activeElementName: "question3",
  currentPageName: "page3",
  pages: {
    page1: {
      passed: true
    },
    page2: {
      passed: true
    }
  }
};

const data = {
  //...
};

survey.data = data;
survey.uiState = prevState;
```
Please let me know if you have any further questions.

nakazonot:
Thanks @JaneSjs, `uiState` solved it — setting `survey.uiState` after `survey.data` marks the preceding pages as passed on resume. Appreciate the help!

PR #11477, opened 2026-06-22
fix(mask): use plain text from paste event for masked inputs

### LINKED ISSUE #11476: Input mask: Masked field incorrectly formats a value when `beforeinput` supplies HTML
T27503 - Pasting into masked telephone input displays unexpected numbers on Microsoft Edge
https://surveyjs.answerdesk.io/internal/ticket/details/T27503

JaneSjs:
Some browsers supply HTML clipboard content via beforeinput event.data when pasting into an input, which causes pattern masks to extract digits from markup instead of the copied phone number. Read text/plain from the paste event and prefer it over beforeinput data for insertFromPaste.

### DIFF
--- a/packages/survey-core/src/mask/input_element_adapter.ts
+++ b/packages/survey-core/src/mask/input_element_adapter.ts
@@ -4,6 +4,7 @@ import { ITextInputParams } from "./mask_utils";
 export class InputElementAdapter {
   private prevUnmaskedValue: string = undefined;
   private maskedEmptyValue : string;
+  private pastedPlainText: string | undefined = undefined;
 
   private setInputValue(value: string) {
     if (this.inputElement.maxLength >= 0 && this.inputElement.maxLength < value.length) {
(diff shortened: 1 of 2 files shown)


JaneSjs:
Closing without merge.

The user's logs showed that on a normal paste, beforeinput (insertFromPaste) received HTML from the clipboard instead of plain text, while text/plain on the clipboard was correct. "Paste as plain text" worked as expected.

For masked text inputs, SurveyJS follows the W3C Input Events specification. When pasting into an <input> field, the beforeinput event with inputType: "insertFromPaste" is expected to provide the inserted content as plain text through event.data. Our input mask implementation uses event.data to format the pasted digits according to the configured mask pattern.

Intercepting the paste event and reading text/plain from the clipboard instead would interfere with the standard browser paste pipeline. Such a selective override is difficult to apply consistently across all environments: it may help in one non-standard setup, but could behave unexpectedly in another. For that reason, we do not plan to change the core mask implementation for this case.

Workarounds for affected environments: use "Paste as plain text", or apply an application-level paste handler.
```js
survey.onAfterRenderQuestionInput.add((sender, options) => {
  const question = options.question;
  if (!question.maskType || question.maskType === "none") return;

  const input = options.htmlElement;
  if (input.dataset.maskPasteFixAttached) return;
  input.dataset.maskPasteFixAttached = "true";

  input.addEventListener("paste", (e) => {
    e.preventDefault();

    const plain = e.clipboardData.getData("text/plain");
    const masked = question.maskInstance.getMaskedValue(plain);
    input.value = masked;

    const unmasked = question.maskInstance.getUnmaskedValue(masked);
    if (unmasked) {
      question.value = question.maskSettings.saveMaskedValue ? masked : unmasked;
    }
  });
});
```

PR #11574, opened 2026-07-14
Guard window event-listener helpers when window lacks addEventListener

### LINKED ISSUE #11573: `survey-core` throws an exception during import in React Native: `window.addEventListener` is not a function
## Description

`survey-core` registers a `touchmove` listener on `window` at **module load time**, in the drag-and-drop DOM adapter. The guard only checks that `window` exists — it does not check that `window.addEventListener` exists:

```js
// packages/survey-core/src/dragdrop/dom-adapter.ts (compiled form, survey.core.js)
if (DomWindowHelper.getWindow()) {
  DomWindowHelper.getWindow().addEventListener("touchmove", function (event) {
    if (!DragDropDOMAdapter.PreventScrolling) return;
    event.preventDefault();
  }, { passive: false });
}
```

React Native defines a global `window` object (in recent versions it's literally aliased to `global`), but that object has **no `addEventListener` method**. The `if (DomWindowHelper.getWindow())` check passes because `window` is truthy, and the call throws immediately:

```
TypeError: ... .addEventListener is not a function
```

Because this runs at **module load** — before any component renders, before `Survey.settings` can be configured, and before any `document`/DOM APIs are touched — this crashes even purely headless (model-only, no UI) usage of `survey-core` in React Native. There's no way to work around it from application code: the import itself throws.

## Framework-free repro (plain Node, no React Native needed)

React Native's `window` shape can be approximated in plain Node — no device, emulator, or RN toolchain required:

```js
global.window = {};      // window exists but has no addEventListener, like React Native
require('survey-core');  // throws
```

Captured output against a fresh build of `master`:

```
TypeError: DomWindowHelper.getWindow(...).addEventListener is not a function
    at .../survey-core/build/survey.core.js:33587:37
    at .../survey-core/build/survey.core.js:8:68
    at Object.<anonymous> (.../survey-core/build/survey.core.js:11:3)
    at Module._compile (node:internal/modules/cjs/loader:1730:14)
    ...
```

## Why this happens in React Native specifically

React Native's JS runtime (Hermes) exposes a global `window` for compatibility with libraries that check `typeof window !== "undefined"`, but that object is not a real DOM `Window` — it has no `addEventListener`/`removeEventListener`. Any library that treats "`window` exists" as "full DOM API is available" will hit this. This is the same failure signature as [TanStack/query#155](https://github.com/TanStack/query/issues/155) ("React Native: `window.addEventListener is not a function`"), which was a bug in a different library with the identical root cause.

## Relation to existing issues

- #7218 / #7629 — a previous React Native (Windows) crash was fixed by guarding an unavailable global (`globalThis`, falling back to `self`) in `settings.ts`. This is the same *class* of bug — an unguarded assumption about global object shape in a non-browser JS environment — but a different call site (`globalThis` vs. `window.addEventListener`), so it wasn't fixed by that change.
- #7327 — open architectural issue asking to reduce module-load-time side effects and global-scope access in `survey-core`. This bug is a concrete instance of exactly that pattern (a DOM listener registered as a side effect of importing the module), so a fix here is in scope of that effort.
- #6713 — meta-issue tracking React Native support requests. React Native is acknowledged as unsupported/untracked in general, but this specific crash blocks even the simplest headless use (building `Model` and reading/writing values with no UI at all), so it's worth tracking and fixing independently of full RN support.

I searched the tracker (open and closed) for `touchmove`, `addEventListener is not a function`, `DomWindowHelper`, `PreventScrolling`, and `React Native` + `dragdrop`/`window.addEventListener`, and didn't find an existing report of this specific crash.

## Suggested fix

`global_variables_utils.ts`'s `DomWindowHelper` already exposes guarded `addEventListener`/`removeEventListener` wrappers, but their guard only checks `typeof window !== "undefined"`, not that `addEventListener` is actually a function on it — so the wrappers have the same gap as the raw call in `dom-adapter.ts`. Strengthening the wrappers' guard (and routing the `touchmove` registration through them, since it currently bypasses them via `DomWindowHelper.getWindow().addEventListener(...)`) fixes this at the shared DOM-access layer rather than as a one-off patch. I have a minimal, behavior-preserving fix ready and will open a PR referencing this issue.

## Environment

- `survey-core`: reproduced against a build of `master` (package.json version `2.5.33` at time of writing)
- Runtime: React Native (Hermes) — `window` is defined without `addEventListener`; also reproduced in plain Node by shadowing `global.window = {}`

markholland:
## Title
Guard window event-listener helpers when window lacks addEventListener

## Description

Fixes #11573

`DomWindowHelper.addEventListener`/`removeEventListener` (in `packages/survey-core/src/global_variables_utils.ts`) guard only on `typeof window !== "undefined"`. Environments like React Native define a global `window` that has no `addEventListener`/`removeEventListener` methods, so that guard passes and the call throws — most visibly through the drag-and-drop module's `touchmove` listener, which is registered at **module load time** and previously bypassed the helper entirely (`DomWindowHelper.getWindow().addEventListener(...)`), so it crashed even purely headless (no-UI) usage of `survey-core` in React Native. Repro and details in #11573.

### Change

- `global_variables_utils.ts`: `DomWindowHelper.addEventListener`/`removeEventListener` now also check `typeof window.addEventListener === "function"` (resp. `removeEventListener`) before calling it, and accept an optional `options` parameter (`AddEventListenerOptions` / `EventListenerOptions`) forwarded to the underlying call. This is additive — existing call sites that don't pass `options` are unaffected.
- `dragdrop/dom-adapter.ts`: the module-load-time `touchmove` listener registration now goes through `DomWindowHelper.addEventListener(...)` (with `{ passive: false }` passed as `options`) instead of the raw `DomWindowHelper.getWindow().addEventListener(...)` call, so it benefits from the strengthened guard.

Behavior is unchanged everywhere `window.addEventListener` is actually available (verified: same event type and options are passed through). Where it's not available, the previous behavior was an uncaught `TypeError` at import time; the new behavior is a silent no-op, consistent with how every other `DomWindowHelper` method in this file already degrades when `window`/its members aren't available (e.g. `getSelection`, `requestAnimationFrame`, `matchMedia`).

### Testing

- Added/ran the framework-free repro from the issue against a fresh build: `global.window = {}; require('survey-core')` — no longer throws; a headless `Model` can still be constructed and read/written.
- Verified (via a `window` stub with `addEventListener` present) that the `touchmove` listener still registers with the same event type and `{ passive: false }` options as before — no behavior change for browsers/WebViews.
- `npm run build` — clean.
- `npx vitest run tests/dragdropcoretests.ts tests/dragDropMatrixTests.ts tests/dragdrophelpertests.ts` — 29/29 passing.
- `npx vitest run tests/surveytests.ts` (references `DomWindowHelper`) — 733/733 passing.
- `npx eslint src/global_variables_utils.ts src/dragdrop/dom-adapter.ts --max-warnings=0` — clean.

### DIFF
--- a/packages/survey-core/src/dragdrop/dom-adapter.ts
+++ b/packages/survey-core/src/dragdrop/dom-adapter.ts
@@ -6,20 +6,18 @@ import { DomDocumentHelper, DomWindowHelper } from "../global_variables_utils";
 
 // WebKit requires cancelable `touchmove` events to be added as early as possible
 // see https://bugs.webkit.org/show_bug.cgi?id=184250
-if (DomWindowHelper.getWindow()) {
-  DomWindowHelper.getWindow().addEventListener(
-    "touchmove",
-    (event) => {
-      if (!DragDropDOMAdapter.PreventScrolling) {
-        return;
-      }
+DomWindowHelper.addEventListener(
+  "touchmove",
+  (event) => {
+    if (!DragDropDOMAdapter.PreventScrolling) {
+      return;
+    }
 
-      // Prevent scrolling
-      event.preventDefault();
-    },
-    { passive: false }
-  );
-}
+    // Prevent scrolling
+    event.preventDefault();
+  },
+  { passive: false }
+);
 
 export interface IDragDropDOMAdapter {
   startDrag(event: PointerEvent, draggedElement: any, parentElement: any, draggedElementNode: HTMLElement, preventSaveTargetNode: boolean): void;
--- a/packages/survey-core/src/global_variables_utils.ts
+++ b/packages/survey-core/src/global_variables_utils.ts
@@ -45,13 +45,13 @@ export class DomWindowHelper {
       return window.requestAnimationFrame(callback);
     }
   }
-  public static addEventListener(type: string, listener: (e?: any) => void): void {
-    if (!DomWindowHelper.isAvailable()) return;
-    window.addEventListener(type, listener);
+  public static addEventListener(type: string, listener: (e?: any) => void, options?: boolean | AddEventListenerOptions): void {
+    if (!DomWindowHelper.isAvailable() || typeof window.addEventListener !== "function") return;
+    window.addEventListener(type, listener, options);
   }
-  public static removeEventListener(type: string, listener: (e?: any) => void): void {
-    if (!DomWindowHelper.isAvailable()) return;
-    window.removeEventListener(type, listener);
+  public static removeEventListener(type: string, listener: (e?: any) => void, options?: boolean | EventListenerOptions): void {
+    if (!DomWindowHelper.isAvailable() || typeof window.removeEventListener !== "function") return;
+    window.removeEventListener(type, listener, options);
   }
 
   public static matchMedia(mediaQueryString: string): {matches:boolean} | null {


markholland:
Superseded by https://github.com/surveyjs/survey-library/pull/11575

PR #11655, opened 2026-07-30
fix: dependency-skip misses case-insensitive references to dynamic pa…

### LINKED ISSUE #11653: Flag constant-vs-constant comparisons as `SemanticError` in `SurveyModel.validateExpressions`
T27766 - FormPath — a browser-local checker for SurveyJS conditional logic
https://surveyjs.answerdesk.io/internal/ticket/details/T27766

In condition expressions (`visibleIf`, `enableIf`, `requiredIf`, etc.), report a `SemanticError` when both sides of `=` `==` / `!=` are literals (e.g. 'property-use' == 'residential'). Such rules never depend on answers and are almost always authoring mistakes from free-text input.

MarchTokenda:
…nel context variables

Expression variables are resolved case-insensitively ({parentpanel.q1} matches the parentPanel context variable), but the changed-keys check used by the visibleIf/enableIf/requiredIf dependency-skip compared names case-sensitively. A question inside a nested dynamic panel with visibleIf "{parentpanel.q1} ..." was not re-evaluated when the parent panel value changed.

- getContextKeys: register lower-case aliases for context variable names
- getFirstNameByKeys: fall back to a lower-case key lookup per path step

### DIFF
--- a/packages/survey-core/src/conditions/conditionProcessValue.ts
+++ b/packages/survey-core/src/conditions/conditionProcessValue.ts
@@ -140,6 +140,8 @@ export class ValueGetter {
     for (let i = 0; i < path.length; i++) {
       res += (i > 0 ? "." : "") + path[i].name;
       if (keys.hasOwnProperty(res)) return res;
+      const lowerRes = res.toLowerCase();
+      if (lowerRes !== res && keys.hasOwnProperty(lowerRes)) return lowerRes;
     }
     return "";
   }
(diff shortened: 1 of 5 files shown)


JaneSjs:
@MarchTokenda would you also consider the case when we use other placeholders, such as `row` -> `ROW` in a matrix ?
```
{
  "pages": [
    {
      "name": "page1",
      "elements": [
        {
          "type": "matrixdropdown",
          "name": "question13",
          "columns": [
            {
              "name": "Column 1",
              "title": "YES to show Column 2",
              "cellType": "boolean"
            },
            {
              "name": "Column 2",
              "cellType": "text",
              "visibleIf": "{ROW.Column 1} = true"
            }
          ],
          "choices": [
            1,
            2,
            3,
            4,
            5
          ],
          "rows": [
            "Row 1"
          ]
        }
      ]
    }
  ]
}
```
Please refer to a full list of placeholders which can be used in expresions:
https://surveyjs.io/form-library/documentation/design-survey/conditional-logic#question-values

JaneSjs (approved):

andrewtelnov (approved):

PR #11659, opened 2026-07-31
[creator:7920] Enhance expression validation for constant conditions

MarchTokenda:
… fix: 11653


andrewtelnov:
@MarchTokenda I would add `addConditionSemanticErrors` function into Operand and as result in executor you will have the following code:
```ts
if (options.semantics && isCondition) {
  this.operand.addConditionSemanticErrors(errors);
}
```

andrewtelnov (approved):

PR #11786, opened 2026-08-30
Feature/mask date localization

andrewtelnov:


packages/survey-core/src/localization/czech.ts
```diff
@@ -246,7 +246,13 @@ export var czechSurveyStrings = {
...
-};
-
+  progressbarPage: "Stránka {0}",
+  // "d" - a day placeholder symbol in a datetime input mask
+  maskPlaceholderDay: "d",
+  // "m" - a month placeholder symbol in a datetime input mask
+  maskPlaceholderMonth: "m",
+  // "y" - a year placeholder symbol in a datetime input mask
+  maskPlaceholderYear: "r"
+};
```
novikov82, line 268: what about hours, minutes, seconds?

andrewtelnov: Added the localization strings

novikov82 (approved):

PR #11804, opened 2026-09-03
Fix composite header and content padding not collapsing in panelless mode

marvin-wtt:
## Summary

- In panelless themes (`isPanelless: true`), composite questions (`.sd-panel`, `.sd-composite`, `.sd-question--table`) kept a fixed horizontal inset on their header and content area, even though the panel's own outer box padding correctly collapsed to `0`.
- Root cause: `--sjs2-layout-component-panel-header-padding-left/-right` and `--sjs2-layout-component-panel-content-area-padding-horizontal` are never zeroed for panelless mode — only the analogous "simple" tokens used by plain questions were patched in `initDefaultCssVariables()`.
- Extends the existing panelless patch in `initDefaultCssVariables()` (`packages/survey-core/src/themes.ts`) with the missing zeroed tokens, following the same pattern already used for `--sjs2-color-component-panel-default-bg`. This applies to any theme with `isPanelless: true`, including custom Theme Editor themes, not just the bundled presets.
- Nested panels and Panel Dynamic are unaffected — they use separate `-nested-`/`-dynamic-` tokens that already zero correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marvin-wtt:
Here is an example with a compostie question:
image

OlgaLarina:
Thanks for finding this and for the proposed fix.

We will fix the composite question layout in panelless themes, but this change cannot be merged as it is. These variables are shared with a regular panel:

- `--sjs2-layout-component-panel-header-padding-left`
- `--sjs2-layout-component-panel-header-padding-right`
- `--sjs2-layout-component-panel-content-area-padding-horizontal`

In a panelless theme a panel is still rendered with a frame (`sd-panel--with-frame`). Setting these tokens to zero removes the header and content padding inside that frame, so a normal panel loses its inner inset.

A composite question should be rendered the same way as a panel: the same frame, header separator, and inner padding.

Tracked in https://github.com/surveyjs/survey-library/issues/11895.

PR #11830, opened 2026-09-11
Survey header renders an empty logo image when `logo` contains dynamic text that resolves to an empty value

marvin-wtt:
### Problem

A survey whose `logo` is a placeholder — `{ "logo": "{logo}" }` — renders ``
when the variable is empty or `null`, showing a broken-image glyph and the survey title as alt
text where the logo should be.

### Cause

`hasLogo` was computed from the raw property text (`!!this.logo`), which is always truthy for a
placeholder, while the header takes the image's `src` from `locLogo.renderedHtml` — the
*processed* text. Visibility and `src` were derived from two different values, so they could
disagree. All four renderers share this pattern, so the fix belongs in survey-core.

The `hasLogo` cache was also only invalidated by the `logo` setter and `fromJSON`, with no
subscription to `locLogo.onStringChanged`.

### Fix

- `hasLogo` now reads the same processed value the header renders into `src`.
- Reset the cached value on `locLogo.onStringChanged`, mirroring the existing `locTitle` /
  `titleIsEmpty` subscription.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marvin-wtt:
@andrewtelnov The failing checks look unrelated to this PR. I merged master into this branch, but I cannot run the CI again since I do not have permissions. Please let me know if you'd like any changes.

marvin-wtt:
@andrewtelnov I'm not asking for a timeline, but is this something you would consider merging in the future, or should I instead look for a solution that doesn't rely on using a variable in the logo property?

andrewtelnov:
@marvin-wtt I asked @novikov82 to take a look. He promised to do it.

Thank you,
Andrew

PR #11869, opened 2026-09-17
Select a Rating value with number keys

### LINKED ISSUE #11863: Select a Rating value with number keys
Related issues
#9272 Keyboard input for Rating Scale and Radiogroup questions

Add another way to answer a Rating Scale: press the number that matches the value (`3`, `0`, `10`). Click and keyboard movement along the scale stay as they are. Number keys are an extra commit path, not a replacement.

Always on. No new survey or question property.

When the rating is focused, a number key selects the matching value in one gesture — same commit class as Enter, not as arrows.

This does not conflict with WCAG 2.1.4 or the ARIA radio-group keyboard: digit shortcuts are extra keys, active **only while the rating has focus**. Do not intercept Tab, arrows, or Space.

### In

- `type: "rating"`: `labels`, `stars`, `smileys`
- Auto-generated scales (`rateMin` / `rateMax` / `rateStep` / `rateCount`) and custom `rateValues`
- Rating as a matrix cell (focused cell is a `QuestionRatingModel`)
- Classic layout and `questionsOnPageMode: "questionPerPage" | "inputPerPage"`
- `autoAdvanceEnabled`
- Same behavior in every UI package; logic in `survey-core`

### Out

- Any new API, JSON property, or `settings` key
- Shortcut badges, “press Enter ↵”, OK confirm, changing Enter/arrow behavior
- Dropdown rating, including `displayMode: "auto"` after it switches to dropdown: do not intercept digits

## Behavior

Extend the existing `QuestionRatingModel.onKeyDown`. Do not add a second document-level listener. Call `preventDefault` / `stopPropagation` only when a digit actually selected an item.

Handle digits only when the event target belongs to this rating (focus is on the scale). Ignore when:

- read-only / disabled / design / preview
- dropdown mode
- focus is in a comment or other text input
- Ctrl / Meta / Alt
- a popup is open
- IME (`keyCode === 229`)
- key repeat (one commit per press)

Several ratings on one page: a digit affects only the focused rating.

### Digit mapping

Main keyboard and numpad, against `visibleChoices` / `visibleRateValues` (after `visibleIf` and the smiley cap of 10).

1. **Match by value.** Key `N` selects the item whose numeric value equals `N` (same `parseFloat` path as `valueToData`).

   | Scale | Key | Result |
   | --- | --- | --- |
   | default 1–5 | `3` | `3` |
   | 0–10 | `0` | `0` |
   | 0–10 | `7` | `7` |
   | `rateMin: 1`, `rateMax: 10` | `1` then `0` quickly | `10` |
   | stars / smileys 1–5 | `4` | `4` |

2. **Two-digit values** (10, and 11–20 up to `settings.ratingMaximumRateValueCount`). If a visible value is longer than one digit and starts with the key, wait **300–400 ms**:
   - `1` + `0` inside the window and `10` exists → `10`
   - otherwise after timeout → `1`
   - if there is no prefix candidate (scale 1–5), select immediately
   - a lone `0` never means `10`; it selects value `0` only

3. **Index fallback.** If no value equals the digit (non-numeric `rateValues`, or `rateStep: 2` → `1, 3, 5, 7, 9` and the user pressed `2`): 1-based index over visible items (`1` = first, `9` = ninth). In this mode `0` selects nothing. If nothing matches, ignore and do not `preventDefault`.

4. **Same digit again.** Do not `clearValue` (unlike a second click in `setValueFromClick`). Follow Enter: keep the value and, if `autoAdvanceEnabled`, still auto-advance (`tryGoNextPageAutomatic` when the value did not change).

5. **`enableIf: false` on an item.** The key does not select it, same as a click.

RTL does not affect digits.

### Auto-advance

Reuse the Enter commit flag (`isEnterKey`) or the same pattern: number keys are a commit, arrows are not.

```ts
supportAutoAdvance(): boolean {
  return this.isMouseDown === true
    || this.isDropdown
    || this.isEnterKey === true;
}
```

`setValueFromClick` already bails out while `isEnterKey` is set so Enter does not toggle-clear. Digit handling must not go through `setValueFromClick` either.

Use existing `settings.autoAdvanceDelay`. If auto-advance is off, only write the value.

### Focus

Leave or move focus to the selected item so AT announces the new state. Do not shift vertical layout in `inputPerPage`.

## Implementation notes

- Unit-test the “key → item” helper with no DOM (value match, `10` prefix window, index fallback).
- UI packages already call `question.onKeyDown`; they should not reimplement the digit map.
- Set `question.value` the same way Enter does, not via a synthesized click.
- Prefer `aria-keyshortcuts` on the item control in this PR if it does not drag in adapters; otherwise a follow-up. Do not put the shortcut into the accessible name.
- Because this is always on, intercept digits **only** while the rating is focused (WCAG 2.1.4). Never steal keys from comments or a dropdown filter.

## Edge cases

- `rateStep !== 1`: value first, then index.
- Smileys truncated to 10: shortcuts only for visible items.
- `visibleIf` on `rateValues`: a hidden item does not consume a digit and cannot be selected.
- Empty scale: ignore.
- `showCommentArea`: shortcut still sets the rating; digits go to the textarea only when that field is focused.
- Dynamic panel / matrix: target the focused rating, not the first one on the page.
- Survey completed or navigating: handler is inert.

## Acceptance criteria

- [ ] No new survey/question property
- [ ] Click, arrows, Enter, and Space are unchanged, including arrows without auto-advance and Enter with auto-advance
- [ ] Labels / stars / smileys 1–5: with focus on the rating, `3` → `3` in `survey.data`
- [ ] Scale 0–10: `0` → 0, `7` → 7, `1`+`0` quickly → 10, lone `1` after timeout → 1
- [ ] Non-numeric `rateValues`: `1` / `2` / `3` by index
- [ ] Dropdown rating: digits are not intercepted
- [ ] Number key + `autoAdvanceEnabled` advances (with delay), including when that value was already selected; `ArrowRight` does not
- [ ] Focus in comment: digits type into the field; rating does not change
- [ ] Several ratings on one page: a digit affects only the focused rating
- [ ] `readOnly` / disabled / preview: ignore
- [ ] Works in every UI package
- [ ] Existing e2e `check auto next page with rating + keyboard` (arrows + Enter) keeps its meaning

### Unit (`question_ratingtests.ts`)

- 1–5, 0–10 with the `10` timer, index fallback, `rateStep: 2`
- `supportAutoAdvance`: arrows / number key / Enter / mouse / dropdown
- Repeating the same digit does not clear the value; with auto-advance it still advances
- `readOnly`, dropdown: no-op
- Existing `supportAutoAdvance on Enter, bug#11845` still passes

### E2E

- Focus the rating, press `4`, assert survey result
- Auto-advance: digit advances, arrow does not
- Focus in comment: digit goes into the text

## Open questions

1. Disambiguation window for `10`: 300 vs 400 ms; treat 11–20 the same when the value is two digits (proposal: yes).
2. Single-question modes (`questionPerPage` / `inputPerPage`): handle digits only when focus is inside the rating (same as classic), or also when the current element is that rating and focus is not in a text field? Proposal: same rule as Enter today — only via the existing `onKeyDown` on the scale.

OlgaLarina:


packages/survey-core/src/question_rating.ts
```diff
@@ -84,6 +84,52 @@ export class RatingItem extends ItemValue {
   }
 }
 
+export const ratingDigitShortcutDelay = 400;
```
andrewtelnov: Let's move it into the survey settings. @RomanTsukanov please pick up the correct name
It make apply for other questions as well.

andrewtelnov (approved):

PR #11882, opened 2026-09-18
validate() does not report a value a question cannot hold, and clearI…

### LINKED ISSUE #11876: `validate()` does not report a value a question cannot hold, and `clearIncorrectValues` can throw
## Summary

When a `SurveyModel` is used headlessly to check a stored response on a server, the payload did not come from an editor, so it can contain values no editor produces: a string in a matrix, an object in a text question, a choice that is not in the list, a row key that no question owns.

Against **`survey-core` 3.1.0** the documented flow crashes on one such input and silently accepts many others:

```js
const model = new Model(json);
model.data = payloadFromAnUntrustedClient;
model.clearIncorrectValues(true);   // throws on some inputs, keeps most of the bad ones
const ok = model.validate(false, false);   // true
```

The root cause: **nothing tells whether a question can hold the value it has.** `Question.isNewValueCorrect` is `return true` for every type except the container types, a value it rejects stays in `survey.data` anyway, and `validate()` never looks at any of it.

## Resolution

Instead of teaching `clearIncorrectValues` more cases and asking the server to call it, the question itself answers "is my value correct?", and `validate()` reports the answer. The server flow becomes:

```js
const model = new Model(json);
model.data = payloadFromAnUntrustedClient;
const ok = model.validate(false, false);   // false for a value a question cannot hold
```

`clearIncorrectValues` is not needed for the check, and **the check does not modify the data it was given**.

### New API

- **`Question.isValueCorrect(): boolean`** — tells whether the question can hold the value it has: the value has the JSON shape the question stores, and it refers to existing choices, rows and items only. It reads the question value and the raw value in the survey data (a container does not take `"oops"`, but the survey keeps it). It never modifies either.
- **`IncorrectValueError`** — `getErrorType()` is `"incorrectvalue"`, the text is the new `incorrectValueError` localization string ("The value is incorrect."). `validate()` adds it to a question whose `isValueCorrect()` is `false`.
- **`clearIncorrectValues()`** is built on the same check: a question with an incorrect value clears it. Checkbox, the matrices, dynamic panel and multiple text still remove only the incorrect part (an unknown choice, row, key or item) and keep the rest.

For question type authors, two protected functions:

- `isDataValueCorrect(val)` — **renamed from `isNewValueCorrect`**. Decides whether the question takes a value at all, on assignment and from `survey.data`. It stays lenient on purpose: a dropdown has to take a value before its choices are loaded.
- `isValueCorrectCore(val)` — the full check behind `isValueCorrect()`; calls `isDataValueCorrect` first and adds the rules that must not block assignment.

### What is reported as incorrect

| Question | Incorrect value |
|---|---|
| `text`, `comment` | an object; for `inputType` `number`/`range` a value that is not a number; for `date`, `datetime-local`, `month` a value that is not a date |
| `boolean` | anything except `valueTrue` and `valueFalse` |
| `radiogroup`, `dropdown`, single `imagepicker` | an array; a value that is not a choice |
| `checkbox`, `tagbox`, `ranking`, multi `imagepicker` | an item that is not a choice |
| `rating` | a value that is not a rate value |
| `file` | anything except a file entry or an array of them |
| `imagemap` | a value that is not an area; an array in single mode and a scalar in multi mode |
| `matrix` | not an object; an unknown row; a cell value that is not a column |
| `matrixdropdown` | not an object; an unknown row; a row that is not an object |
| `matrixdynamic`, `paneldynamic` | not an array; a row or panel that is not an object; a key that no column or question stores |
| `multipletext` | not an object; an unknown item |

A cell, a panel question and a multiple text item are questions too, so a bad nested value is reported by the nested question.

The choices check follows the rules `clearIncorrectValues` already had: it is off for `keepIncorrectValues` (survey or question), lazy-loaded choices, `choicesByUrl` that has not answered, a carry-forward source that is not ready, and custom choices. Keys are resolved by `valueName`: questions that share a `valueName` know the rows and items of each other, a detail panel question is found by its `valueName`, and comment (`-Comment`) and totals keys are known keys.

### Behavior notes

- **`validate(false)` is a pure check.** A select question used to call `clearIncorrectValues()` inside every validation. It still does when errors are shown (`fireCallback` is `true`): a respondent cannot fix a value whose choice is gone, so it is removed. A silent `validate(false)` reports the value and leaves the data as it is.
- The check is skipped for validation on value changed: a value set through the UI is correct by construction.
- An array given to `text`/`comment` becomes a string and a scalar given to `checkbox` becomes a one-item array. These are existing, deliberate conversions and are not reported.
- A `matrixdropdown` row that is not an object is reported, but not refused on assignment; `clearIncorrectValues` removes such rows.

## Scenarios of the original report

| # | Scenario | Status |
|---|---|---|
| 1 | `clearIncorrectValues` throws on `matrixdynamic` rows that are not objects | **Fixed.** Such an array is not taken by `matrixdynamic` and `paneldynamic`, is reported by `validate()` and removed by `clearIncorrectValues` |
| 2 | A value the question does not take stays in the data and passes `validate()` | **Fixed.** Reported by `validate()`, removed by `clearIncorrectValues` |
| 3 | The warning is malformed and printed twice | **Partly.** The missing space is fixed; the warning is still printed more than once per assignment |
| 4 | Scalar questions accept objects and arrays | **Fixed**, except an array in `text`/`comment`, see the notes above |
| 5 | `inputType` is not validated | **Fixed** for `number`, `range`, `date`, `datetime-local`, `month`. `time`, `week`, `email` and `url` are not checked |
| 6 | Containers are checked unevenly | **Partly.** `multipletext` reports and clears unknown items. "An emptied matrix row deletes the matrix value, an emptied panel stays as `{}`" is unchanged |
| 7 | `ranking` rewrites its value on assignment | **Not addressed**, needs a separate issue |
| 8 | Limits the UI enforces (`max*`, `maxLength`, `readOnly`) are not validation rules | **Not addressed**, needs a separate issue |

## Also fixed on the way

- `clearIncorrectValues` of a `matrix` or `matrixdropdown` removed the rows of another question sharing its `valueName`.
- `clearIncorrectValues` of a matrix row erased the answers of detail panel questions when the detail panel was not created yet or a question had a `valueName`, and erased `-Comment` values of cell questions.
- The unique-column check of `matrixdropdown` counted invisible rows (the `Bug#10395.2` test did not catch it because its data had rows that were not objects).

## Compatibility

- `validate()` can now fail where it passed: for a value outside the choices, an unknown row, item or key, a value of a wrong shape. Set `keepIncorrectValues` to turn the choices check off.
- `isNewValueCorrect` is renamed to `isDataValueCorrect`. It is protected; a custom question class that overrides it has to rename the override.

andrewtelnov:
…ncorrectValues can throw, fix #11876


andrewtelnov:
[New PR](https://github.com/surveyjs/survey-library/pull/11896)

PR #11884, opened 2026-09-19
Issue/11680 incorrect 50 percent width 2

novikov82:


Voyagerroc-Lab (approved):
Reviewed with assistance from OpenAI Codex. The element wrapper now stops growing when an explicit width is supplied on desktop, while mobile keeps the available-width behavior. Min/max constraints move to the element root, and the row spacing/width selectors now apply consistently to page and panel rows. The updated core tests and UI snapshots cover those layout changes.

